PR #13, opened 2019-11-08
Split the tests into multiple files/classes for clarity

tarkatronic:
Having a single `test_all.py` isn't a great pattern, especially with a single test class inside it. This splits the tests out to one test file per actual code file. This gives them a more logical place to live, makes them more reportable, and easier to browse.


irodelta (approved):
looks good

Issue #18, opened 2019-11-15
When cloning a remote repo, config should be read from its pyproject.toml


tarkatronic:
We will soon be able to read default configuration from a local `pyproject.toml` file, but this leaves a bit of a gap:

When you invoke `tartufo` with a `git_url`, it will clone that url into a local folder, and then scan that folder. But if that repo contains a `pyproject.toml` with a `[tool.tartufo]` section, we don't respect any of those options.

I don't think it makes sense, in this scenario, to read *all* options. But I think it does make sense to scan *select* options. The options I think we should attempt to read are:
* `rules` -- files containing regexes
* `include-files` -- Files to be included in the scan
* `exclude-files` -- Files to be excluded from the scan

PR #19, opened 2019-11-15
Read config from pyproject.toml, and use `click` for CLI parsing

### LINKED ISSUE #14: Allow configuration via file
To help with guaranteed repeatability, it should be possible to configure `tartufo` by way of a configuration file, as opposed to always having to specify options on the command line.

My personal recommendation would be to add the configuration to the `pyproject.toml` file exclusively, in a `[tool.tartufo]` section, as with [black](https://github.com/psf/black#pyprojecttoml). But other standard places which would be acceptable include:
* `setup.cfg` in a `[tartufo]` section
* `.tartufo.yml`

Some tools allow configuring from 3 distinct locations. I think that, since we are starting fresh, a single location would be ideal. And given that `pyproject.toml` is becoming The Way for configuration, that is my recommendation.

The goal is that all command line options should be configurable in this file, or files, such that simply running `tartufo` will always produce the same results, with the desired options.

tarkatronic:
Fixes #14 

This ended up being a bit more ambitious than intended, but in a very good way. Many things shook out of this change, and it's brought up a few other issues and made more progress possible, as well as made things more testable.

* You can now set defaults in a local `pyproject.toml` file, in a `[tool.tartufo]` section. Note that we do not (yet) read this file from an automagically cloned repo. That ended up being a bit of a rabbit hole, and somewhat more complex. This is tracked in #18.
* All command line options have been standardized to use `-` as word separators. Previously, some had `_` and some had `-`.
* `tartufo` now uses a PEP-518 build system specification, since we have a `pyproject.toml`
* There is a lot more test coverage now!
* The configuration/option checks are short-circuited a bit better now
* There should be no more bombing out to the command line with a stack trace
* All of the "boolean" options are now flags, and all have negating forms. For example, `--json` and `--no-json`.
* Requirements got loosened up; they were previously pinned to specific versions, which is not great.

There is probably more, but that's what I remember for right now!


tartufo/cli.py
```diff
@@ -1,200 +1,138 @@
...
+              help="Scan staged files in local repo clone.")
+@click.option("--config",
+              type=click.File(mode='r'),
+              is_eager=True,
+              callback=config.read_pyproject_toml,
+              help="Read configuration from specified file. [default: pyproject.toml]")
+@click.argument("git_url", required=False)
+@click.pass_context
+def main(ctx, **kwargs):
+    """Find secrets hidden in the depths of git.
```
mxhenry-godaddy, line 67: Why no Type-ing now?

mxhenry-godaddy (approved):

tartufo/cli.py
tarkatronic: Good catch, I totally missed that!

irodelta (approved):

Issue #23, opened 2019-11-21
The searchOrg script should become a first class citizen


tarkatronic:
## Is your feature request related to a problem? Please describe.

Right now, `searchOrg.py` exists under the `scripts` folder as a completely isolated entity. As the code evolves, this is guaranteed to break, because it's not currently testable. It's also not ideal to run, since you have to manually install extra dependencies, which are not documented.

Also, the inclusion of all the regex rules in the script is somewhat conspicuous, since all others have been moved out to an external module.

This current structure is just not maintainable.

## Describe the solution you'd like

The `searchOrg` script should be migrated, either to a `console_script`, how `tartufo` itself is, or moved out to an external package. Something like `tartufo-org-scanner`.

## Describe alternatives you've considered

The only other option I can see would be to leave it where it is and try to keep it up to date in-place. But this is less than ideal, for reasons mentioned up in the problem section.

## Teachability, Documentation, Adoption, Migration Strategy

Given that there are is no current documentation for this script, I don't think there's any "migration path", per se. We would just be implementing a new entry point; a whole new feature.

rscottbailey:
Well, in the spirit of reducing the scope of the problem, i suggest beginning by gutting the entire regex definition section. I scanned it quickly by eye and the list appears identical to the one in truffleHogRegexes.

At this point, we are down to an iterator that sweeps everything within an organization (hardcoded to those hosted by github). Actually, I see the organization name is hardcoded also, to "square"...

If we think that project enumeration is quasi-portable (say to GHE), then i would suggest just grafting it into tartufo proper (i.e. `--organization foobar --url https://something.or.other/blah/blah/blah` assumes you mean all projects in organization foobar on `https://something.or.other` ignoring remainder of URL) and pretty much what we move over to `../tartufo` is the enumeration logic.

If that logic is not going to work anywhere except github.com itself, then I'd probably just write the script off as a curiosity and circular-file it before somebody gets into trouble with it.

tarkatronic:
I actually had some thoughts on this while I was driving earlier. There have been some things bothering me about this entire code base, including passing around so many different parameters, and the return values being these amorphous dictionaries. And some duplicated code scattered around. And I think I came up with an overall solution which could also snag this problem up. Allow me to tell the rest of this tale in some pseudo code.

```python
class BaseScanner:
    def scan_blob(self, blob):
        if self.do_regex:
            self.regex_scan(blob)
        if self.do_entropy:
            self.entropy_scan(blob)

[...]

PR #24, opened 2019-11-22
Read config files from remote repos

### LINKED ISSUE #18: When cloning a remote repo, config should be read from its pyproject.toml

tarkatronic:
Fixes #18.

I added a few things in here:

* We now look for `tartufo.toml` in addition to `pyproject.toml`, for non-Python projects
* When scanning a remote git repo, we look for config files and load their values for `include-paths` and `exclude-paths`. These struck me as the most vital to load from the repo config. Others can be added as necessary, but this seemed like a good first effort.
* Moved the git clone functionality outside the `find_strings` function, because it absolutely did not belong in there in the first place.
* Introduced a new `scan_repo` function which gives a single code path for both remote and local repos being scanned. It is called *after* the git clone, performs the config lookups, and then passes on to `find_strings`.
* Marked the project as "typed", or fully type annotated, so mypy can treat it as such

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -371,7 +371,7 @@ def scan_repo(
         if extra_paths:
             file_path = pathlib.Path(extra_paths).resolve()
             if file_path.is_file():
-                with file_path.open() as paths_file:
+                with file_path.open("r", encoding="utf8") as paths_file:
                     path_inclusions.extend(
                         config.compile_path_rules(paths_file.readlines())
                     )
@@ -379,7 +379,7 @@ def scan_repo(
         if extra_paths:
             file_path = pathlib.Path(extra_paths).resolve()
             if file_path.is_file():
-                with file_path.open() as paths_file:
+                with file_path.open("r", encoding="utf8") as paths_file:
                     path_exclusions.extend(
                         config.compile_path_rules(paths_file.readlines())
                     )
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -361,7 +361,7 @@ def scan_repo(
     if not config_file.is_file():
         config_file = path / "tartufo.toml"
     if config_file.is_file() and str(config_file.resolve()) != str(options["config"]):
-        toml_file = toml.load(config_file)
+        toml_file = toml.load(str(config_file))
         repo_config = toml_file.get("tool", {}).get("tartufo", {})
     if repo_config:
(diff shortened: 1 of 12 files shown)


mxhenry-godaddy (approved):

tartufo/scanner.py
```diff
@@ -343,11 +342,61 @@ def find_strings(
             branch_name,
         )
         output = handle_results(output, output_dir, found_issues)
-    output["project_path"] = project_path
-    output["clone_uri"] = git_url
+    output["project_path"] = repo_path
```
rscottbailey, line 345: I'm generally a fan of this cleanup, but do we assume that the caller will be able to identify/guess the remote repository URL from the generated scratch path? That is, are we losing useful information by no longer stashing git_url in the findings?

rscottbailey (approved):
I made one comment inline but I don't think it's an issue at this point since I believe we can only specify a single target repository per invocation and therefore the source of any findings is unambiguous. It's probably something to think about if we ever drag the organization stuff back in or otherwise enable the caller to accumulate outputs from multiple repository scans.

tartufo/scanner.py
tarkatronic: That is a good call out. My main check was to see if these were used anywhere else in this code base, and they are not. I didn't think about this being included in the actual json files which get written out, and whether it would be useful there. @mxhenry do you know if these values are currently used by any of our tooling?

I think this would be good to include in our refactoring of issues/output to a class-based structure spawned via #23.

mxhenry-godaddy: Replied in email

tarkatronic: Glancing through our current usage, it doesn't look like we're using those pieces right now. I think it is useful to have though, so we should worry about including it in the 2.0 refactoring.

PR #29, opened 2019-12-09
Represent issues with a class, rather than a dictionary

tarkatronic:
Taking the text of the primary commit in this PR:

> This was a rather large shift in the code, and the start of a larger cleanup discussed, in part, in #23. The actual changes include:
> 
> * `Issue` is now a top level class used to represent issues found in scanning The `Bcolors` class is gone, replaced by the styling functionality of1 the `click` library
> * `print_results` has been removed, in favor of the `__str__` method on the `Issue` class. This greatly improves the testability of the code and makes it easier to output everything at once and potentially even use a pager for output in the future.
> * JSON output now uses the `as_dict` method on the `Issue` class, for a more clear representation of what exactly is output.
> 
> There is also one larger implied change here; I have not been explicitly including Python 2.7 support. The one immediate place this is likely to be a problem is in the use of the `enum` library. This was not added to Python until 3.4. While there is a backport available via the `enum34` package, I hesitate to add this. I would prefer to take this as the beginning of 2.0 development, in which we drop support for Python 2.7 entirely.

The one notable deviance, since writing that, is that Python 2.7 is still supported. It turns out that `enum34` is easy enough of a drop-in backport to not use it. For now. This does not, however, change my opinion that we should drop Python 2 support in the next major version. But I think that this should be done as one atomic PR, as opposed to piecemeal as this would have been.

Also, it was far too easy to adapt the tests to the new code. I changed one single assertion, even after I made massive changes to the internal API. This makes my confidence in the reliability of the tests even lower. I think they need some pretty major reworking before they can be truly relied upon. But that should come naturally with the rest of the work in #23.

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -5,18 +5,20 @@ from __future__ import division
 from __future__ import print_function
 
 import datetime
+import enum
 import hashlib
 import json
 import math
 import os
-import sys
 import tempfile
 import uuid
 from typing import cast, Dict, Iterable, List, Optional, Pattern, Set, Union
 
+import click
 import git
 import toml
 from tartufo import config
+from tartufo.util import style_ok, style_warning
 
 try:
     import pathlib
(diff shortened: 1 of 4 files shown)


rscottbailey (approved):

tartufo/scanner.py
```diff
@@ -32,19 +34,108 @@
...
+        if not self.commit:
+            return None
+        return self.commit.message
+
+    @property
+    def commit_hash(self):
+        # type: () -> Optional[str]
+        if not self.commit:
+            return None
+        return self.commit.hexsha
```
mxhenry-godaddy, line 106: Is it possible to do ternary here?
```return None if not self.commit else self.commit.hexsha```?

File: tartufo/scanner.py [hunk 2]
```diff
@@ -32,19 +34,108 @@
...
+    RegEx = "Regular Expression Match"
+
+
+class Issue:
+    """Represents an issue found while scanning the code."""
+
+    OUTPUT_SEPARATOR = "~~~~~~~~~~~~~~~~~~~~~"  # type: str
+    DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S"  # type: str
+
+    issue_type = None  # type: Optional[IssueType]
```
mxhenry-godaddy, line 48: What is the advantage of this vs
```IssueType issue_type = None``` for the declaration?

**Review by mxhenry-godaddy — APPROVED** [review]

File: tartufo/scanner.py (hunk 2 above)
Comment (tarkatronic) [reply tartufo/scanner.py:48]: Well, the advantage of this way is that it's valid Python syntax. :wink: 
The proper inline annotation in Python would be:
```python
issue_type: IssueType = None
```
As described in [PEP 526](https://www.python.org/dev/peps/pep-0526/#abstract). But, this particular syntax is only available in Python 3.6 and above, meaning we would have to drop Python 3.5 support in order to use this syntax. A move like this is not entirely unprecedented, for example in the [black](https://github.com/psf/black) autoformatter. But given that Python 3.5 is still [officially supported until 2020-09-13](https://devguide.python.org/#branchstatus) I'm a little bit hesitant to just go ahead and drop that support now. But of course, that's not off the table. :)

tartufo/scanner.py
tarkatronic: It is possible to do that, yes. I've just always found that syntax to be rather less readable. While it is certainly more compact, I find it less easy to mentally parse.

PR #31, opened 2019-12-10
Drop support for Python 2

### LINKED ISSUE #30: Drop Python 2 support

Python 2.7 is officially EOL as of Jan 1, 2020. And we have to go through all kinds of effort to keep things working against it. We should explicitly stop that.

Down with Python 2.7! Remove the future imports! Remove the backports!

And since we'll then be Python 3.5+, we can start using [PEP 484 style type hints](https://www.python.org/dev/peps/pep-0484/#type-definition-syntax) for function signatures, instead of all the comments. We will have to keep the comments for variables though, until such time as we exclusively support 3.6+.

The alternative is... not doing it. Which is just not so much an option.

This will be a new major version, so it will be easy to say that version 2.0+ only supports Python 3.5+, and if Python 2.7 support is needed, then users will have to stick to 1.x.

tarkatronic:
Python 2 is dead! Long live Python 3!

This removes all support for Python 2, and start using some more recent syntax. From here on, only 3.5+ are supported.

* Removed all `__future__` imports
* Removed the backport packages `enum34`, `mock`, `pathlib2`, and `typing`.
* Stopped testing against Python 2
* Moved the function type annotations inline (inline variable annotations are only available in 3.6+)

Fixes #30

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
 language: python
 python:
-  - "2.7"
   - "3.5"
   - "3.6"
   - "3.7"
   - "3.8"
-  - "pypy"
   - "pypy3"
 install: pip install tox-travis codecov
 script: tox
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ regex = true
 entropy = true
 
 [tool.black]
-target-version = ['py27', 'py35', 'py36', 'py37', 'py38']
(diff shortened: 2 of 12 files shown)


rscottbailey (approved):
As usual, just looking at the differences makes a great primer/refresher for python best coding practices. ;-)

mxhenry-godaddy (approved):

PR #33, opened 2019-12-13
Fix docs & pre-commit hook to use hyphens in CLI arguments

tarkatronic:


rscottbailey (approved):
Looks good. I also did a check to see if I could spot any more instances that should have been changed but weren't, and it looks to me like you caught everything.

PR #36, opened 2019-12-13
Reverse sense of difference

### LINKED ISSUE #32: Diff output shows additions as deletions and vice versa
## 🐛 Bug Report

It appears that the diff output from `tartufo` is produced in the wrong direction; lines which were added in a commit are shown as deleted, and lines which were deleted in the commit are shown as added. For example:

```diff
~~~~~~~~~~~~~~~~~~~~~
Reason: High Entropy
Filepath: tartufo/scanner.py
Branch: origin/master
Date: 2019-12-10 12:34:56
Hash: 2b46d80fcff3defb102dd0a1b844c7cbb521767a
Commit: Remove backport imports

@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-

+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
 import datetime
 import enum
 import hashlib
 import json
 import math
 import os
-import pathlib
 import tempfile
 import uuid
 from typing import cast, Dict, Iterable, List, Optional, Pattern, Set, Union
@@ -17,6 +20,11 @@ import toml
 from tartufo import config
 from tartufo.util import style_ok, style_warning

+try:
+    import pathlib
+except ImportError:
+    import pathlib2 as pathlib  # type: ignore
+

 BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
 HEX_CHARS = "1234567890abcdefABCDEF"

~~~~~~~~~~~~~~~~~~~~~
```

As you can see from the commit message, this actually *removed* the backward compatibility imports, but this shows that they were *added* in this commit.

I tested this in `truffleHog` as well, and it exhibits the same behavior.

## To Reproduce

From the root of a `tartufo` clone, run the following:

```sh
tartufo --repo-path . --since-commit 16312b6935d98cfb24c9493dcdfa85593196f065 --cleanup
```

## Expected Behavior

The diff should show code alterations as they actually happened, rather than reversed.

## Environment

I have the following configuration in my `pyproject.toml` (the same which is checked into the `tartufo` repo):

```toml
[tool.tartufo]
repo-path = "."
json = false
cleanup = true
regex = true
entropy = true
```

rscottbailey:
Fixes #32.

All of the other difference producers used the correct ordering.

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -315,7 +315,7 @@ def find_strings(
             if not prev_commit or diff_hash in already_searched:
                 prev_commit = curr_commit
                 continue
-            diff = prev_commit.diff(curr_commit, create_patch=True)
+            diff = curr_commit.diff(prev_commit, create_patch=True)
             # avoid searching the same diffs
             already_searched.add(diff_hash)
             found_issues = diff_worker(
--- a/tests/test_scanner.py
+++ b/tests/test_scanner.py
@@ -57,7 +57,7 @@ class ScannerTests(unittest.TestCase):
         )
 
         call_1 = mock.call(
-            commit_1.diff.return_value,
+            commit_2.diff.return_value,
             None,
             True,
             False,
@@ -69,7 +69,7 @@ class ScannerTests(unittest.TestCase):
             master_branch.name,
         )
         call_2 = mock.call(
-            commit_2.diff.return_value,
+            commit_3.diff.return_value,
             None,
             True,
             False,


tarkatronic:
@rscottbailey Thanks for the patch! Looks like the tests need to be updated to match this. If you can take care of that, we can get this merged in. :-)

rscottbailey:
I added a second commit that makes the unit tests work again.

tarkatronic (approved):

Issue #44, opened 2020-02-25
Running tartufo from docker does not scan target repo


mxhenry-godaddy:
## 🐛 Bug Report

Running Tartufo from the docker image in the recommended manner results in Tartufo scanning itself in the docker image, not the target repository

## To Reproduce

> docker pull godaddy/tartufo
> docker run -t --rm -v "$PWD:/git" godaddy/tartufo --repo_path .

results in tartufo scanning the tartufo directory in the container, not the mounted repo.

Changing the invocation to reference the mountpoint correctly fixes the scan target issue, but leads to failing ssh

> docker run -t --rm -v "$PWD:/git" godaddy/tartufo --repo-path /git
> Traceback (most recent call last):
> ...
>   cmdline: git fetch -v origin
>   stderr: 'error: cannot run ssh: No such file or directory

## Expected Behavior

Docker container should be able to execute the above command successfully and without user interaction

## Code Example
At minimum requires the addition of 
> apk add openssh-client

to the Dockerfile

janerikcarlsen:
Added https://github.com/godaddy/tartufo/pull/45 to make the docker image work properly for pre-commit hooks. 
This PR does not solve ssh-agent authentication to github repositories for full history scan.

c-jcarlsen-godaddy:
PR merged, suggest closing this issue

tarkatronic:
Resolved in #45 & #46

PR #45, opened 2020-03-09
Change working directory to avoid tartufo self-scan

janerikcarlsen:
This is a fix for the pre-commit part of [Issue #44](https://github.com/godaddy/tartufo/issues/44) when using the docker image for tartufo secret scanning.

Changing working directory from `/app` go the mounted directory `/git` to avoid tartufo scanning itself. The `--repo-path` will then default to `.` which will be the current `/git` directory that contains the repository we want to scan. 
 
Usage instructions in a pre-commit scenario: 
```
docker run -t --rm -v "$PWD:/git" godaddy/tartufo --pre-commit
```

However, when omitting the `pre-commit` option, tartufo will attempt to do a full history scan that requires cloning the repository. For this to work with the docker image, we would need to mount ssh keys or use a type of github deploy key with access to the repository onto the docker image when running it. Following up this scenario in a separate PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,5 +6,7 @@ WORKDIR /app
 
 RUN pip install -e .
 
-ENTRYPOINT [ "tartufo" ]
-CMD [ "-h" ]
+WORKDIR /git
+
+ENTRYPOINT [ "tartufo"]
+CMD ["-h"]
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -32,11 +32,10 @@ in a local repository. The repository location can be specified using
 caller's current working directory is assumed to be somewhere within the local
 clone's tree and the repository root is determined automatically.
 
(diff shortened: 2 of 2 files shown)


rxmarbles (approved):

tarkatronic:
Thanks for the PR!

This fix looks good for the specific use case where it's failing. What I would like to see though, is documentation of the Docker usage, in the `docs/` folder! Right now, we offer a Dockerfile, but it's left to the user to figure out how to employ that.

That could be a separate PR though. :-)

janerikcarlsen:
Thanks for reviewing. 

I added instructions for running the docker image as pre-commit hook. 

For the other modes of operation (without the --pre-commit option), those are not fully supported in the docker image yet. The documentation for those modes should follow the PRs containing the implementations.

sjacoby-godaddy (approved):

PR #46, opened 2020-03-20
Add SSH Keys auth to docker image and update docs

janerikcarlsen:
This PR aims to solve the remaining part of https://github.com/godaddy/tartufo/issues/44 where ssh keys are needed to clone history for private repositories. 

- Added openssh-client to `Dockerfile`
- Added scripts and configs to mount ssh keys with or without passhprase to ssh-agent 
- Updated documentation to cover docker usage


tarkatronic:
@janerikcarlsen Thank you for continuing to work on this! Give me a little time to review this; my Docker-fu is a bit rusty, but something is itching at the back of my brain suggesting there might be a slightly cleaner way to do this.

I'll also take a look at what's going on with Travis... it seems to be getting errors right now with configuring pytest, which is odd. Doesn't seem to be your fault though! 😁

tarkatronic:
For my part, I am [+0 on this change](https://docs.djangoproject.com/en/3.0/internals/contributing/bugs-and-features/#how-we-make-decisions). I'm not super up-to-date on Docker best practices, so I'm not comfortable giving a full +1 here. I would like to lean on @beardedeagle for such a review.

sjacoby-godaddy (approved):

docs/source/usage.rst
```diff
@@ -23,6 +30,24 @@ without the need for the temporary clone:
 
    $ tartufo --repo-path /path/to/my/repo
 
+For ``docker``, mount the local clone to the ``/git`` folder in the docker image:
+
+.. code-block:: sh
+
+   $ docker run --rm -v "/path/to/my/repo:/git" godaddy/tartufo 
+
+When scanning private repositories, the ``docker`` image needs to have access to SSH keys for authorization. Mount the SSH key to ``/tmp/id_rsa`` in the docker image:
```
silasbw: How about forwarding the SSH agent via the volume (and env vars)?  Something like:

```
docker run --rm -v $SSH_AUTH_SOCK:/agent -e SSH_AUTH_SOCK=/agent ...
```

That will remove the requirement from passing in the key password, means you don't need to run the agent in the container, etc

It adds the requirement that local machine/host needs to be running ssh-agent.

c-jcarlsen-godaddy: Thanks for the suggestion. PR is updated to use ssh-agent forwarding instead of passing keys and passphrase

silasbw (approved):

tarkatronic (approved):
I love it! This is great. I'm going to get this merged in, and I am also going to be working on getting all of the recent fixes backported to the 1.x branch, so that we can have a new stable release.

Thank you @janerikcarlsen and @silasbw!

PR #49, opened 2020-04-20
Add a doc for example cleanup process walkthrough

sjacoby-godaddy:


mxhenry-godaddy (approved):

tarkatronic (changes requested):
@sjacoby-godaddy Thanks for adding this! The "what to do next" is something which we've really needed here.

Before we get this merged, I want to work through a bit of cleanup to make this as easy to manage as possible for our users. Unfortunately I haven't yet gotten around to documenting any style guidelines for writing documentation. For a start, I would encourage two things:

1. If you have access to MacOS X or Linux (including WSL), run: `tox -e docs`. This will build a local copy of the documentation, as it will appear on the final Read The Docs site, under the `docs/build/html/` folder. This file you have created will be named `examplecleanup.html`.
2. Browse around the rest of the documentation to get a feel for the tone of writing. As much as possible, it would be nice to maintain a consistent writing tone to any readers. This should make the docs easier to follow. I suggest this because I would love to see a bit more verbosity on what this document is covering, why, and how.

The most immediate concern I have right now is just cleaning up warnings produced from step 1 above. Those currently look like:
```
docs/source/examplecleanup.rst:2: WARNING: Title underline too short.

tartufo Full Example
=======
docs/source/examplecleanup.rst:5: WARNING: Title underline too short.

End-to-End Example
-----------

checking consistency... docs/source/examplecleanup.rst: WARNING: document isn't included in any toctree
docs/source/examplecleanup.rst:11: WARNING: Could not lex literal_block as "console". Highlighting skipped.
docs/source/examplecleanup.rst:43: WARNING: Could not lex literal_block as "console". Highlighting skipped.
docs/source/examplecleanup.rst:86: WARNING: Could not lex literal_block as "console". Highlighting skipped.
```
Most of these should be fairly self-explanatory. The `toctree` one may not be immediately clear though. What this boils down to is that you need to add a reference to the document under the `toctree` directive at the bottom of `index.rst`.

docs/source/examplecleanup.rst
```diff
@@ -0,0 +1,116 @@
+Would you like to know more?
+============================
+
+End-to-End Example
+------------------
+
```
tarkatronic, line 6: Can you add something up top here, along the lines of:

> In order to follow along with this guide, you will also want to install the `BFG` tool from ...

Just talking about BFG without any kind of introduction is a bit off-putting.

docs/source/examplecleanup.rst
```diff
@@ -0,0 +1,116 @@
...
+An End-to-End example walkthrough of tartufo cleanup
```
tarkatronic: Could you expand this a little bit? Just saying `tartufo cleanup` I think begs the question, "How does this differ from `tartufo --cleanup`?"

Maybe saying something like, "Now that you have found secrets in your repo, what do you do with that information?" ...but, you know, maybe a bit better wording than that. :smile:

docs/source/index.rst
```diff
@@ -18,6 +18,7 @@ Example
 -------
 
 .. image:: _static/img/example_issue.png
+.. _usage:
```
tarkatronic: This shouldn't be here; this makes the link below point up here, instead of to the `usage` page.

docs/source/index.rst
```diff
@@ -59,7 +60,7 @@ Getting started is easy!
       # Mount a local clone of a repo and scan it using docker!
       $ docker run --rm -v "/path/to/your/git/repo:/git" godaddy/tartufo
       
-   For more detail on usage and options, see :ref:`usage`.
+   For more detail on usage and options, see :ref:`Link title <usage>`.
```
tarkatronic: I'm not sure this is quite what you intended to do... This now looks like:
> For more detail on usage and options, see Link title.

docs/source/usage.rst
```diff
@@ -86,7 +86,7 @@ pre-commit hook by using ``git commit --no-verify``.
 If you would like to automate these hooks, you can use either the ``Python`` or ``Docker`` approach to setting up tartufo as a pre-commit hook
 
 Python pre-commit hook
-+++++++++++++++++++++
+++++++++++++++++++++++
```
tarkatronic, line 89: Thanks for fixing this!

docs/source/usage.rst
```diff
@@ -118,6 +118,9 @@ Use the docker image as pre-commit hook by adding the docker run command to ``.g
     docker run -t --rm -v "$PWD:/git" godaddy/tartufo --pre-commit
     EOF
 
+
+.. _examplecleanup:
```
tarkatronic: Similar to the `.. _usage:` on the index, this shouldn't be here; it turns this into an anchor, so that the link below simply points here.

docs/source/usage.rst
```diff
@@ -128,3 +131,6 @@ these files when tartufo completes, specify the ``--cleanup`` flag:
 .. code-block:: sh
 
    tartufo --cleanup
+
+
+Would you like to know more? See :ref:`Link title <examplecleanup>`.
```
tarkatronic: Same as on the index, this now looks like:
> Would you like to know more? See Link title.

tarkatronic (changes requested):
This is great! Just a few more small tweaks -- almost there!

I love that this has become more of a narrative, as opposed to a list of steps.

tarkatronic (approved):
:boom: :microphone: :droplet:

PR #52, opened 2020-05-06
Convert the project from setuptools -> poetry

tarkatronic:
This sets us up with [Poetry](https://python-poetry.org/) instead of setuptools. This way we actually have a legit package manager, dev environment manager, deterministic builds, etc etc etc.

Steps for this to be considered complete:

* [X] Move setup.py metadata into pyproject.toml
* [X] Update sphinx to pull version from pyproject.toml
* [X] Update tox to work with/use Poetry
* [X] Update Dockerfile to work with/use Poetry
* [X] Update build script(s) to work with/use Poetry
* [x] Update Travis config to use Poetry
* [x] Update contributing docs to guide users on using Poetry
* [x] Deprecate `setup.cfg`
  * Move config to `pyproject.toml` where possible
  * Otherwise, move to tool-specific config file
  * Bonus: Bring other config files into `pyproject.coml` where possible

I'm pretty sure that's all that's left... I will add any other things here which pop up.


tarkatronic:
@kquerna I would love to get your input on the updates to the contributing docs here. :smile:

CONTRIBUTING.md
```diff
@@ -1,165 +1,10 @@
 # Contributing
 
-Everyone is welcome to contribute to GoDaddy's Open Source Software.
-Contributing doesn’t just mean submitting pull requests. To get involved you can
-report or triage bugs and participate in discussions on the evolution of each
-project.
+First of all, thank you for considering contributing to our project! We welcome
+contributions of all shapes and size from all contributors!
```
kquerna: Nit: "...shapes and **sizes**"

docs/source/CONTRIBUTING.md
```diff
@@ -126,19 +128,42 @@ automatically on every commit, which is detailed below!
...
+  black: commands succeeded
+  mypy: commands succeeded
+  pylint: commands succeeded
+  vulture: commands succeeded
+  docs: commands succeeded
+  congratulations :)
+$
+```
+
+If you do not have all of the supported Python versions, that's perfectly okay.
```
kquerna: Nit: remove unneeded "of" after all -->  "...do not have all the supported..."

kquerna (approved):

mxhenry-godaddy (approved):

Dockerfile
```diff
@@ -1,14 +1,31 @@
...
+
+RUN apk add --no-cache gcc libffi-dev musl-dev openssl-dev
+RUN pip --no-cache-dir install "poetry==$POETRY_VERSION"
+RUN python -m venv /venv
+
+COPY pyproject.toml poetry.lock ./
+RUN poetry export -f requirements.txt | /venv/bin/pip install -r /dev/stdin
+
+COPY . .
+RUN poetry build && /venv/bin/pip install dist/*.whl
```
bosatsu, line 20: Curious why we are going this route, rather than just using `poetry install` and `poetry build`?

docs/source/CONTRIBUTING.md
```diff
@@ -102,15 +102,17 @@ speaking, Git handles attribution automatically.
...
+to get used to the basic usage.
 
-```sh
-> python -m venv .venv
-> source .venv/bin/activate
-> pip install -e .[tests]
-```
+Now, setting up a development environment is super simple!
+
+Step 1: [Install Poetry]
```
bosatsu, line 112: Consider including a link on how to install poetry?
https://python-poetry.org/docs/#installation

Dockerfile
tarkatronic: The idea here is that we ship a final image with none of the build artifacts; only the finally installed product. If you notice, poetry is not installed into the venv, and so doesn't actually get copied over to the final image. Same with the `-dev` packages installed on line 12. The main goal here is to provide as compact and clean an image as we can.

So with all of that in mind, the easiest way to get Poetry to use the Python interpreter in the venv, is to do this export and pip install.

Of course, all this is adapted from a number of examples I've found online. So I'd be happy if somebody is able to help build a better way!

docs/source/CONTRIBUTING.md
tarkatronic: [Like this](https://github.com/godaddy/tartufo/pull/52/files#diff-c8ea9f02a746ef0a0c7e94b9c291a944R187)? 😄

bosatsu (approved):

Issue #43, opened 2020-02-21
Report out of the list of exclusions


tcostello-godaddy:
## Feature Request

I would like a CLI output and/or a output.txt file generate with the list of exclusions when the scan was run.

## Is your feature request related to a problem? Please describe.

When run the scan with exclusions and it is clean, no output to CLI.

## Describe the solution you'd like

Output to CLI:
date:time
'02/20/2020 - 100% Clean'
List of exclusions:
excludethisfile.txt

tarkatronic:
Would this be just a printout of the patterns which were used for exclusions, or a printout of the actual files which matched those patterns?

I think this makes sense in 2 particular scenarios:

1. Add this to the JSON output. That's pretty much a no-brainer.
2. Add a `-v/--verbose` flag, and this would be the first thing exposed by that.

Issue #42, opened 2020-02-21
Report output on clean or successful scan


tcostello-godaddy:
## Feature Request

Need output to screen and or text file when scan results are clean.

Example:
scan run: 02-20-2020 
scan result: 100% Clean. Success

or

'02/20/2020 - 100% Clean'

## Is your feature request related to a problem? Please describe.

If scan is clear, command line returns nothing. Safe to assume the scan was clean....but would like CLI feedback.

## Describe the solution you'd like

CLI returns when no errors/issues:

'date:time- 100% Clean'

tarkatronic:
I think this makes sense. By default, we can print out some kind of success message.
`All clear. No secrets detected.`
We can also add a `-q/--quiet` flag to suppress that ouput.

This also goes hand in hand with #11 and #37. I think we need generally just a revamp of our outputs.

janerikcarlsen:
Taking this on together with https://github.com/godaddy/tartufo/issues/43

Issue #37, opened 2019-12-16
Remove log location output if no security risks are detected


matcartmill-old:
## Is your feature request related to a problem? Please describe.

No.

## Describe the solution you'd like

By default, when no security risks are discovered Tartufo should not output a log location. This should be opt-in using a `--output-log-location` flag.

## Describe alternatives you've considered

Nothing.

## Teachability, Documentation, Adoption, Migration Strategy

I envision something like `tartufo --regex --output-log-location` to show log location when no risks are found. Otherwise, no output when no risks are found.

tarkatronic:
I'm not sure a new flag is the best solution here. I think our best option is likely to treat this as a bug and simply remove the message if the files are not saved.

tarkatronic:
This got resolved in #82

Issue #11, opened 2019-10-31
Optimize handling of temp files (vs --cleanup)


rscottbailey:
This annoys me so much. ;-)

1. Can we make `--cleanup` the default behavior, at least if no findings are found? It makes no sense that 99.999% of the time the repo is clean and the scratch directory is empty, why should the user have to expend extra effort to remove the empty directory?

Frankly, I'm not sure that --cleanup shouldn't be the default all the time, anyway. Findings are reported on stdout anyway, so the utility here seems to be the case where you expect to have massive numbers of findings and want to save them conveniently for later review.

My proposal would be this:
* eliminate `--cleanup` (both the command flag and the `util.clean_outputs()` function)
* push creation of `output_dir` from scanner's `find_staged()` and `find_strings()` into `handle_results()` where it is done on a just-in-time basis when needed
* add a new flag like `--save-to-dir` which enables calling `handle_results()`, and which defaults to false/none, which means we never call `handle_results()` unless the user asks to save the files, and never create the directory unless we have something to put into it.

2. In addition to, or alternatively, could we at least include "tartufo" in the name (say as a prefix argument to `mkdtemp`)?

decompil3d:
Nit: `-o` is often used in other packages (e.g. aws cli, kubectl) for output format, not output location. Can we consider using something like `--out-dir`/`-od`?

rscottbailey:
Very satisfied customer. ;)

tarkatronic:
This isn't closed quite yet! I'm going to leave this open just until #82 gets merged. Just in case. And then this will auto-close with that PR. 😄

tarkatronic:
@rscottbailey Any chance you can help out with a review of #82 to make sure it works as expected?

PR #57, opened 2020-05-26
Separation of concerns

tarkatronic:
This was work I actually did several months back, as a general code cleanup. As the title and branch names indicate, this is working to achieve a "separation of concerns". This is to hopefully make the code easier to read, follow, debug, and test.

There will likely be more work along these lines along the path to 2.0, but this is currently at a working point and I don't want the PR to get too large.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,7 +1,7 @@
 import unittest
 from unittest import mock
 
-from tartufo import util
+from tartufo import scanner, util
 
 
 class GitTests(unittest.TestCase):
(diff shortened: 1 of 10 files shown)


tests/test_util.py
```diff
@@ -30,3 +30,79 @@ def test_tartufo_clones_git_repo_into_temp_dir(self, mock_mkdtemp, mock_clone):
...
+                        "strings_found": ["bar"],
+                        "commit_time": None,
+                        "commit_message": None,
+                        "commit_hash": None,
+                        "file_path": None,
+                        "branch": None,
+                    },
+                ],
+            }
+        )
```
bosatsu, line 78: Should we add a test for when as_json is false?

tests/test_util.py
```diff
@@ -30,3 +30,79 @@ def test_tartufo_clones_git_repo_into_temp_dir(self, mock_mkdtemp, mock_clone):
...
+
```
bosatsu, line 79: Seems like we should have at least one test for `write_outputs`.

bosatsu (changes requested):

tests/test_util.py
tarkatronic: Right up above. :) https://github.com/godaddy/tartufo/pull/57/files#diff-5832adf374920d75d4bf48e546367f53R36-R41

tests/test_util.py
tarkatronic: Hmmm, good callout on this. Interestingly, the coverage output claims that this function is mostly covered by the tests, even though we aren't explicitly testing it right now.

I think I might want to hold off on writing tests for this until we do an overall reworking of the whole output/logging story. From a quick glance, we've got several issues related to that topic including #37, #42, #43, and possibly also #11. This may even be a case for setting up a new project here on the repo, or at least adding another label for the issues.

bosatsu (approved):

tests/test_util.py
bosatsu: Fair enough I know this is a temp state while trying to get to 2.0

Issue #58, opened 2020-05-29
Convert CI/CD to GitHub Actions


tarkatronic:
## Is your feature request related to a problem? Please describe.

We currently use TravisCI for our builds. It's got a whole host of problems; far too many to go into right now. Suffice to say, we don't want to use it anymore.

## Describe the solution you'd like

All of the current functionality performed by Travis should be replicated by GitHub Actions. As an added bonus, we may be able to get the individual tests steps (different tox environments) listed as separate build checks, making it easier to discover exactly what failed.

## Describe alternatives you've considered

We've looked at CircleCI, but it's had its own set of problems with reliability, and recently on asherah we've had issue with visibility of builds. Plus, GitHub Actions is built right in here, and works *super* well.

## Teachability, Documentation, Adoption, Migration Strategy

Very little should need to be changed in the docs for this. In fact, I'm not sure our CI/CD pipeline is documented at all right now. But we can leave that for another issue...

janerikcarlsen:
Happy to take this on

PR #59, opened 2020-08-07
Convert CI/CD from TravisCI to GitHub Actions

janerikcarlsen:
This PR addresses https://github.com/godaddy/tartufo/issues/58 to convert the build and publish process away from TravisCI over to GitHub Actions. 

There are two workflows added: 
- `ci`
- `publish`

The `ci` workflow triggers on any commit (all branches and forks) and runs tests using tox for python versions `3.5`, `3.6`, `3.7`, `3.8` and `pypy3`. For `3.7` it also runs `black`, `mypy`, `pylint`, `vulture`, and `docs`.
The workflow needs the following secrets created in the repository: 
- `CODECOV_TOKEN` - A token is needed to upload coverage data to codecov. Previously this token might not have been needed, as TravisCI seems to have handled this integration for us (See `codecov --help` for context). Codecov is only uploaded for `python 3.7`

The `publish` workflow is triggered only when a new release is published (not for pre-releases). It will then package using Poetry and upload to PyPi using `poetry publish` as before. The workflow will then build the docker image and publish this to `godaddy/tartufo` in Docker Hub. 
The workflow needs the following Secrets created in the repository: 
- `DOCKER_USERNAME`  - Username with privileges to upload to `godaddy/tartufo` in DockerHub
- `DOCKER_PASSWORD` - Password or API Token for the user with privileges to upload to  `godaddy/tartufo in DockerHub
- `PYPI_TOKEN` - API Token with privileges to publish new versions of `tartufo` in PyPi.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
+name: ci
+
+on: push
```
tarkatronic: This doesn't seem to be working right to trigger the build. These should have caused the action(s) to run on this PR.

I think you're going to want something more like this: https://github.com/Facet-MUD-Project/facetjs/blob/master/.github/workflows/node.js.yml#L6-L10

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  tox:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8, pypy3]
+    steps:
+    - uses: actions/checkout@592cf69
```
tarkatronic, line 12: Why this specific hash? Magic numbers/identifiers like this make me nervous. Can we just use `@v2`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  tox:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8, pypy3]
+    steps:
+    - uses: actions/checkout@592cf69 
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@0c28554
```
tarkatronic, line 14: Ditto the above. I would much rather use a tagged version.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
+name: publish
+
+on:
+  release:
+    types: [created] # Trigger when release is created
```
tarkatronic: Are you sure this is only for when a release is created? Or would this also trigger for other types of "creation" like branch, tag, etc?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@592cf69 
+    - name: Set up Python 3.7
+      uses: actions/setup-python@0c28554
```
tarkatronic, line 13: Ditto on the versions here

.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
...
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run:  |
+        pip install --upgrade pip
+        pip install --upgrade poetry
+    - name: Package and publish with Poetry
+      run: |
+        poetry config pypi-token.pypi $PYPI_TOKEN
+        poetry publish --build
```
tarkatronic: Why `--build`?

tarkatronic (changes requested):
Just a few things here. Looking good so far! Thanks for doing this works, @janerikcarlsen!

.github/workflows/ci.yml
decompil3d: Our policy for 3rd party actions is to target a specific SHA to avoid a tag poisoning attack.

decompil3d (approved):

.github/workflows/publish.yml
janerikcarlsen: According to https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release this should only fire when release actions happen. However I changed the action from `created` to `published` to avoid pre-releases being published.

.github/workflows/publish.yml
janerikcarlsen: I needed `--build` to publish to an alternative registry when testing, removed it from the workflow

.github/workflows/ci.yml
janerikcarlsen: I believe this is expected behavior when creating a PR from a fork - a `pull_request` event is being submitted, and the workflow in the base repo (`godaddy/tartufo`) is ran if it has a matching `on:` statement. Since there are no GitHub Action workflows defined yet in `godaddy/tartufo`, nothing is ran, in line with https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories

I added `pull_request` along with the push event, so that in the future, PR's will get built using the base repository workflow.

mxhenry-godaddy (approved):

tarkatronic (approved):

tarkatronic:
@janerikcarlsen Everything built clean! Thanks for the work on this!
https://github.com/godaddy/tartufo/actions/runs/204374277

PR #61, opened 2020-08-13
Improved exclusions

### LINKED ISSUE #2: Whitelist needs more granularity
ISBAT: define down to the line and column of the file whether to ignore a violation. Current implementation of ignoring an entire file is too broad.

tarkatronic:
Fixes #2 

This introduces a few changes and improvements.

First!
Each detected string now has a "signature", generated as a hash of the string itself and the file it is in. So the output on the command line for an issue now looks like:
Screen Shot 2020-08-13 at 4 53 50 PM
Note the line:
```
Signature: 2a3cb329b81351e357b09f1b97323ff726e72bd5ff8427c9295e6ef68226e1d1
```

Second!
These "signatures" can be excluded from the scan result!
On the command line:
```
$ tartufo -a 2a3cb329b81351e357b09f1b97323ff726e72bd5ff8427c9295e6ef68226e1d1
```
Or in the config file!
```toml
[tool.tartufo]
allow-signatures = [
  "2a3cb329b81351e357b09f1b97323ff726e72bd5ff8427c9295e6ef68226e1d1",
]
```

Third!
You may notice more "issues", or matches, when running tartufo now. This is because they are no longer grouped by file. So every matched string is reported as a separate issue. This was necessary to get the signatures properly generated and reported in any kind of friendly manner.

And.... that's about it, I think.


tarkatronic:
Reviewers may notice that the `pyproject.toml` file got slightly rearranged. Apparently VSCode now has a TOML prettifier, and it magically alphabetized everything for me! So don't freak out too much. That's going to be the majority of the changes there. Aside from all the signatures I added to let the self-run pass.

tartufo/scanner.py
```diff
@@ -90,17 +92,26 @@ def file_path(self) -> Optional[str]:
             return self.diff.b_path
         return self.diff.a_path
 
+    @property
+    def signature(self) -> Optional[str]:
+        if self.file_path:
+            return generate_signature(self.matched_string, self.file_path)
+        return None
```
bitcathedrals, line 99: If this error happens I think a fundamental assumption is broken.

better to throw an exception than smuggle a null up the call chain.

bitcathedrals (approved):
The changes look good. In particular the use of empty lists instead of nulls simplified the code. My one comment is a nit.

The docs are seriously broken as you know. They are especially vital since you use a non-standard package manager.

I reviewed the code manually, and I built the code, and ran the full test suite with no errors. Tox rated the code 10/10. I dont have integration tests so black box testing was not performed.

tartufo/scanner.py
tarkatronic: Yeah, that kept going through my head as I was working on this code. I think the path toward 2.0 is going to involve a lot of cleanup of issues like this. There really should always be a file path. I just want to be able to trace through the code and confidently say that before I write methods like this without that extra check, just in case.

tarkatronic:
I can definitely see how the docs are a bit confusing right now. I think partly that may be just that the branching strategy is not entirely ideal. And in moving toward 2.0 we've changed so very much. The particular problem you ran into, trying to get instructions to set up a build environment, is due to the fact that https://tartufo.readthedocs.io/ defaults to the "stable" docs branch, which is built from our 1.x branch. I think this is the right choice for now, since most users will be on those versions. Unfortunately this means that you have to manually switch over to the "master" version of the docs to get the up-to-date development instructions. And that is not at all obvious or intuitive. I'll put some thought into how to improve this experience.

mxhenry-godaddy (approved):

docs/source/configuration.rst
```diff
@@ -55,6 +55,13 @@ line switch, as seen here:
                                      (default), no Git object paths are excluded
                                      unless effectively excluded via the
                                      --include-paths option.
+     -a, --allow-signatures TEXT     Specify signatures of matches that you
+                                     explicitly want to allow, and mark as okay.
+                                     These signatures are generated during the
+                                     scan process, and reported out with each
+                                     individual match. This option can be
+                                     specified multiple times, to allow as many
+                                     signatures as you would like.
```
beardedeagle: Maybe a nitpick, but `allow-signatures` is somewhat ambiguous to me. `exclude-signatures` would be more explicit, IMO. Otherwise, beyond what was mentioned, looks good to me.

beardedeagle (approved):

docs/source/configuration.rst
tarkatronic: I'd have no problem renaming this if there's a consensus. `allow-signatures` is what we settled on in discussions on #2. Any thoughts here @rscottbailey or @decompil3d ?

decompil3d: Sure, `exclude-signatures` seems fine to me.

rscottbailey: I'm fine with `exclude-signatures` as well.

PR #64, opened 2020-08-16
move type annotations out of comments, bump minimum python version, c…

### LINKED ISSUE #62: Remove support for Python 3.5
Python 3.5 is about to hit EOL.
Python 3.6 is better and generally available.
We already enforce >= 3.6 for `aws-okta-processor`.
Python 3.6 has the shiny happy new dictionary implementation.
Python 3.6 has `blake2` built in, so one fewer external dependency.
Python 3.6 has class variable annotations. We can stop using the comments for type annotations. Yay!

Let's do this.

voldemortensen:
…loses #62

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,4 +27,3 @@ jobs:
       uses: codecov/codecov-action@v1
       with:
         file: ./coverage.xml
-        token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,9 @@ jobs:
     - name: Install dependencies
       run:  pip install poetry tox codecov tox-gh-actions
     - name: Run tox environments
-      run: tox -- --cov --cov-append --cov-report=xml
+      run: tox
     - name: Run codecov
       if: matrix.python-version == '3.7'
       uses: codecov/codecov-action@v1
       with:
-        file: ./coverage.xml
+        file: ./.coverage
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8, pypy3]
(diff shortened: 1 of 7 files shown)


tarkatronic:
Hey @voldemortensen, thanks for this work; I'm a little jealous you're getting to do this! 😄 This looks amazing and I'm excited to see this go through. I am hoping to wait until #61 is merged before doing this, because it will also allow the removal of an extra dependency I had to add in there (`pyblake2`). Also of note is that the dependency for `black` can be fixed in this PR. Since we will not be supporting Python < 3.6, we no longer have to specify installing that only on >= 3.6!

tarkatronic:
@voldemortensen You've also exposed a shortcoming in the current CI workflow. I know this is unrelated to this particular branch, but would you mind fixing it up? Either here or in another branch would work. It looks like the coverage information never gets sent to Codecov, because you do not have access to the secrets in this repo. It should be possible to solve this by using the `codecov` action, rather than manually installing and running `codecov` like we currently do. You can see an example of how this works here: https://github.com/Facet-MUD-Project/facetjs/blob/master/.github/workflows/node.js.yml#L33-L36

You would just need to change the path to the actual coverage report file. :)

Thanks!

voldemortensen:
@tarkatronic this was a fun one! I'm in no hurry, so waiting for #61 is fine with me.

I've removed the black version restriction and moved over to the codecov action. I'm not sure way the codecov check hasn't passed since the file has been uploaded. Also, it looks like the repo settings are configured to expect the 3.5 test, so that will need to be updated.

tarkatronic:
It occurred to me, this morning, that I was being silly and my logic was completely backwards. This should absolutely go in before #61 so I never have to introduce that new dependency into the main branch. 😄

tarkatronic (changes requested):
This looks great! Just one last thing: If you can run `poetry update` to sync the `pyproject.toml` with the `poetry.lock` file, we can get this merged!

voldemortensen:
@tarkatronic done!

tarkatronic:
@voldemortensen Perfect, thank you! 🙌  I'm not sure what's with the codecov integration not reporting right as a status update... but I'm going to call this good anyhow.

tarkatronic (approved):

Issue #39, opened 2020-01-17
Split functionality into sub-commands?


tarkatronic:
## Is your feature request related to a problem? Please describe.

This results from some thoughts I've had on a few issues:

1. The difference between specifying `git_url` and `--repo-path` is nebulous at best. Especially since you can specify a filesystem path as `git_url`, and it will just re-clone that to a temporary directory.
2. Neither of those options applies when using `--pre-commit`.
3. We've talked, in #23, about integrating the `searchOrg` functionality into core, as opposed to a hacky script. Adding another `--github-org-url` is... not great.

## Describe the solution you'd like

The main solution I have come up with is to use sub-commands. So then we would end up with, for example:

```sh
$ tartufo scan-remote <git url>
$ tartufo scan-local <filesystem path>
$ tartufo pre-commit
$ tartufo scan-github-org
```

Most options would stay top-level/global, so available to all commands. So you would specify for example:

```sh
$ tartufo --regex --entropy --json scan-local ./my/awesome/repo
```

## Describe alternatives you've considered

The only alternative I can think is to split this into multiple commands... which kinda sucks.

## Teachability, Documentation, Adoption, Migration Strategy

This would definitely be a breaking change, and so would require a major version change. Luckily, this is already planned for 2.0, with a bunch of other breaking changes. We also now have a Read The Docs site set up, which will help with documentation. I've also just found a [sphinx-click plugin](https://github.com/click-contrib/sphinx-click) which might help us automagically document the usage. But I haven't used it yet, so I can't be absolutely sure.

This would be a huge change, so I would love to have some discussion on the topic, and see if this even makes sense, before anybody just dives into it.

PR #65, opened 2020-08-21
Handle config via a dataclass

tarkatronic:
This is the first step in a major code cleanup working toward v2.0.

It is loosely based on the discussion in #23, and is a big step in the work toward #39.

Basically now instead of dissecting and passing parts of a dictionary around piecemeal, the config is all handled through these new `Options` dataclasses. The idea is that, when #39 is implemented, each type of scanner will implement its own new `Options` dataclass, hence the split between `BaseOptions` and `GitOptions`.

This also resulted in a very nice reduction of LOC. A net of just over -100 lines!


beardedeagle:
looking

beardedeagle:
Cursory glance looks good, digging deeper before giving all clear.

beardedeagle (approved):
LGTM

janerikcarlsen (approved):
The change looks great 🔥

Issue #12, opened 2019-11-06
Support scanning bare repos


tarkatronic:
Currently, attempting to scan a bare git repo, using `--repo_path`, produces an error along the lines of:
```pytb
Traceback (most recent call last):
  File "/home/jwilhelm/Documents/workspace/tartufo/.venv/bin/tartufo", line 11, in <module>
    load_entry_point('tartufo', 'console_scripts', 'tartufo')()
  File "/home/jwilhelm/Documents/workspace/tartufo/tartufo/cli.py", line 58, in main
    path_exclusions=path_exclusions,
  File "/home/jwilhelm/Documents/workspace/tartufo/tartufo/scanner.py", line 287, in find_strings
    for curr_commit in repo.iter_commits(branch_name, max_count=max_depth):
  File "/home/jwilhelm/Documents/workspace/tartufo/.venv/lib64/python3.7/site-packages/git/objects/commit.py", line 278, in _iter_from_process_or_stream
    finalize_process(proc_or_stream)
  File "/home/jwilhelm/Documents/workspace/tartufo/.venv/lib64/python3.7/site-packages/git/util.py", line 332, in finalize_process
    proc.wait(**kwargs)
  File "/home/jwilhelm/Documents/workspace/tartufo/.venv/lib64/python3.7/site-packages/git/cmd.py", line 414, in wait
    raise GitCommandError(self.args, status, errstr)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git rev-list --max-count=1000000 1/head --
  stderr: 'fatal: bad revision '1/head'
```
This is because a bare repo is a wholly different structure from a normal clone, and produces different results from `git` operations. The specific problem causing the error here is this:
```pycon
>>> repo = git.Repo('tartufo.git')
>>> for branch in repo.remotes.origin.fetch():
...     print(branch.name)
... 
master
split_tests
1/head
10/head
11/head
16/head
19/head
2/head
20/head
21/head
22/head
23/head
24/head
25/head
3/head
8/head
9/head
v0.0.1
v0.0.2
>>> repo = git.Repo('../tartufo')
>>> for branch in repo.remotes.origin.fetch():
...     print(branch.name)
... 
origin/master
origin/split_tests
>>>
```
All of the `X/head` references are not actual valid git revisions, and so `tartufo` chokes on them.

We should figure out a way to either scan all refs, or only scan actual branches.

sjacoby-godaddy:
[...]

PR #67, opened 2020-09-01
Rebuild the scanner(s) as class(es).

tarkatronic:
**IMPORTANT UP FRONT NOTE:** This is a part of an incremental change. None of this new code is used yet. ~Honestly, it's not even unit tested yet. I would be open to writing unit tests against it before this gets merged.~ I wanted to at least get this PR in front of eyes quickly though.

This is a huge change. It's basically the first step in a major rewrite. This is an attempt to make the code more readable, more extensible, maintainable... and honestly, for some of it, just to understand what it's even doing.

Basically what I'm offering here is a re-implementation of the scanner, in a class-based form. You will notice that there is a class hierarchy here. It currently consists of the following:

```
ScannerBase
|-- GitRepoScanner
```

The goal is that this can be extended, and we can have many more scanner types in the future. I'm envisioning something like:

```
ScannerBase
|-- FileScanner
|   |-- FolderScanner
|-- GitScanner
    |-- PreCommitScanner
    |-- GitRepoScanner
    |   |-- LocalRepoScanner
    |   |-- RemoteRepoScanner
    |-- GithubOrgScanner
```

That complete structure is, of course, WAY down the line. But these can each be implemented one-by-one, and theoretically much more easily. In this way, this code is a large step toward #23. Also, as a direct result of this structure and modularity, this is a major step toward #39 and even #12. Basically, this restructured code enables the future of this project.

I tried to reduce complexity in as many places as I could, but I'm sure I did not succeed everywhere. One place that is evident is in the `GitRepoScanner._iter_branch_commits()` and `GitRepoScanner._iter_diff_index()` methods. They handle some of the nitty gritty of traversing back through git commit history. Previously this had been handled, along with a lot more, in the `find_strings()` function. This is my attempt to make more purpose-built functions, and keep the functions from growing too large. I hope that comes across well.

If you want to try this code out locally, you can do so! This little snippet will get you along your way. It utilizes a utility function I built for tests to build up a full set of options. This is an example of running this class-based scanner against `tartufo` itself:

```pycon
>>> from tartufo.scanner import GitRepoScanner
>>> from tests.helpers import generate_options
>>> from tartufo.types import GitOptions
>>>
>>> sc = GitRepoScanner(generate_options(GitOptions, exclude_paths=open(".tartufo-excludes", "r"), entropy=True, regex=True, exclude_signatures=(), default_regexes=True), ".")
>>>
>>> issues = sc.scan()
>>>
>>> len(issues)
243
>>>
```

Note that I have not yet verified that this 243 number is in any way correct. It may not be. Did I mention there are no unit tests for this new code yet? The accuracy will shake out in those tests.

I hope y'all like this new direction. Let me know.

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -121,7 +121,7 @@ class Issue:
 
 
 class ScannerBase(abc.ABC):
-    _issues: List[Issue]
+    _issues: Optional[List[Issue]] = None
     _included_paths: Optional[List[Pattern]] = None
     _excluded_paths: Optional[List[Pattern]] = None
     _rules_regexes: Optional[Dict[str, Pattern]] = None
@@ -132,7 +132,7 @@ class ScannerBase(abc.ABC):
 
     @property
     def issues(self) -> List[Issue]:
-        if not self._issues:
+        if self._issues is None:
             self._issues = self.scan()
         return self._issues
 
@@ -220,7 +220,8 @@ class ScannerBase(abc.ABC):
                 issues += self.scan_entropy(chunk)
             if self.options.regex:
                 issues += self.scan_regex(chunk)
-        return issues
+        self._issues = issues
+        return self._issues
 
     def scan_entropy(self, chunk: Chunk) -> List[Issue]:
         issues: List[Issue] = []
@@ -267,12 +268,10 @@ class GitRepoScanner(ScannerBase):
     __repo: git.Repo
     options: GitOptions
 
-    def __init__(  # pylint: disable=super-init-not-called
-        self, options: GitOptions, repo_path: str
-    ) -> None:
-        self.options = options
+    def __init__(self, options: GitOptions, repo_path: str) -> None:
         self.repo_path = repo_path
         self.load_repo(self.repo_path)
+        super().__init__(options)
 
     def load_repo(self, repo_path: str):
         self.__repo = git.Repo(repo_path)
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -255,7 +255,8 @@ class ScannerBase(abc.ABC):
                     issues.append(issue)
         return issues
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def chunks(self) -> Generator[Chunk, None, None]:
         """Yield "chunks" of data to be scanned.
 
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -244,11 +244,11 @@ class ScannerBase(abc.ABC):
 
     def scan_regex(self, chunk: Chunk) -> List[Issue]:
         issues: List[Issue] = []
-        for key, pattern in self.rules_regexes:
+        for key, pattern in self.rules_regexes.items():
             found_strings = pattern.findall(chunk.contents)
             for match in found_strings:
                 # Filter out any explicitly "allowed" match signatures
-                if self.signature_is_excluded(match, chunk.file_path):
+                if not self.signature_is_excluded(match, chunk.file_path):
                     issue = Issue(IssueType.RegEx, match)
                     issue.issue_detail = key
                     issues.append(issue)
@@ -345,7 +345,7 @@ class GitRepoScanner(ScannerBase):
                     )
 
             # Finally, yield the first commit to the branch
-            diff = diff_index.diff(git.NULL_TREE, create_patch=True)
+            diff = curr_commit.diff(git.NULL_TREE, create_patch=True)
             for blob, file_path in self._iter_diff_index(diff):
                 yield Chunk(
                     blob,
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -344,13 +344,14 @@ class GitRepoScanner(ScannerBase):
                     )
 
             # Finally, yield the first commit to the branch
-            diff = curr_commit.diff(git.NULL_TREE, create_patch=True)
-            for blob, file_path in self._iter_diff_index(diff):
-                yield Chunk(
-                    blob,
-                    file_path,
-                    util.extract_commit_metadata(prev_commit, remote_branch),
-                )
+            if curr_commit:
+                diff = curr_commit.diff(git.NULL_TREE, create_patch=True)
+                for blob, file_path in self._iter_diff_index(diff):
+                    yield Chunk(
+                        blob,
+                        file_path,
+                        util.extract_commit_metadata(prev_commit, remote_branch),
+                    )
 
 
 def shannon_entropy(data: str, iterator: Iterable[str]) -> float:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -304,7 +304,7 @@ class GitRepoScanner(ScannerBase):
                     since_commit_reached = True
                 if since_commit_reached:
                     prev_commit = curr_commit
-                    continue
+                    break
             if not prev_commit:
                 prev_commit = curr_commit
                 continue
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -125,6 +125,7 @@ class ScannerBase(abc.ABC):
     _issues: List[Issue]
     _included_paths: Optional[List[Pattern]] = None
     _excluded_paths: Optional[List[Pattern]] = None
+    _rules_regexes: Optional[Dict[str, Pattern]] = None
     options: GlobalOptions
 
     def __init__(self, options: GlobalOptions) -> None:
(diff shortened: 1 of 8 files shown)


tartufo/scanner.py
```diff
@@ -90,35 +91,269 @@ def file_path(self) -> Optional[str]:
...
+    options: GlobalOptions
+
+    def __init__(self, options: GlobalOptions) -> None:
+        self.options = options
+
+    @property
+    def issues(self) -> List[Issue]:
+        if self._issues is None:
+            self._issues = self.scan()
+        return self._issues
```
bosatsu, line 137: Just a comment based on a discussion with @tarkatronic  that this may end up not being expressed as `@property`, as it may work better to populate this from the `ScannerBase.scan()` method.  The implementation of this will most likely shed light on what makes more sense.

tartufo/scanner.py
```diff
@@ -90,35 +91,269 @@ def file_path(self) -> Optional[str]:
...
+                self.options.default_regexes,
+                self.options.rules,
+                self.options.git_rules_repo,
+                self.options.git_rules_files,
+            )
+        return self._rules_regexes
+
+    @lru_cache()
+    def should_scan(self, file_path):
+        """Check if the a file path should included in analysis.
```
bosatsu: `should *be* included`

tartufo/scanner.py
```diff
@@ -90,35 +91,269 @@ def file_path(self) -> Optional[str]:
...
+
+        If non-empty, `self.included_paths` has precedence over
+        `self.excluded_paths`, such that a file path that is not matched by any
+        of the defined `self.included_paths` will be excluded, even when it is
+        not matched by any of the defined `self.excluded_paths`. If either
+        `self.included_pats` or `self.excluded_paths` are undefined or empty,
```
bosatsu: `self.included_pat*h*s`

tartufo/scanner.py
```diff
@@ -90,35 +91,269 @@ def file_path(self) -> Optional[str]:
...
+                    if not self.signature_is_excluded(string, chunk.file_path):
+                        hex_entropy = self.calculate_entropy(string, HEX_CHARS)
+                        if hex_entropy > 3:
+                            issues.append(Issue(IssueType.Entropy, string))
+        return issues
+
+    def scan_regex(self, chunk: Chunk) -> List[Issue]:
+        issues: List[Issue] = []
+        for key, pattern in self.rules_regexes.items():
+            found_strings = pattern.findall(chunk.contents)
```
bosatsu, line 249: I'm curious if it's possible for a bad regex to come through the process and for this to fail because `pattern` is not a valid regex.  It didn't seem like `utils.configure_regexes()` or `rules_regexes()` are checking for this.

bosatsu (changes requested):
Finished review of `ScannerBase` and `test_base_scanner.py`.   Will review the rest hopefully tomorrow.

tartufo/scanner.py
tarkatronic: If there is a bad regex, it should fail pretty quickly during when the first item is scanned. `util.configure_regexes` calls `util.load_rules_from_file`, which subsequently calls `re.compile` on all of the patterns. If a bad regex it found, the user will see an error along the lines of:

```pycon
>>> import re
>>> re.compile("bad(regex")
  File "lib/python3.7/re.py", line 234, in compile
    return _compile(pattern, flags)
  File "lib/python3.7/re.py", line 286, in _compile
    p = sre_compile.compile(pattern, flags)
  File "lib/python3.7/sre_compile.py", line 764, in compile
    p = sre_parse.parse(p, flags)
  File "lib/python3.7/sre_parse.py", line 924, in parse
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
  File "lib/python3.7/sre_parse.py", line 420, in _parse_sub
    not nested and not items))
  File "lib/python3.7/sre_parse.py", line 813, in _parse
    source.tell() - start)
re.error: missing ), unterminated subpattern at position 3
>>>
```
I think this is absolutely desired behavior. We don't want to silently pass on invalid regex patterns.
Or are you suggesting that we wrap this exception and raise our own?

tartufo/scanner.py
tarkatronic: The inspiration behind this is the way that forms work in Django. Specifically, when a user submits data to a form, you can have code along the lines of:

```python
form = MyForm(data=request.POST)
if form.errors:
    # Handle errors
```
In accessing `form.errors`, it will automatically "clean" the data for you by calling `self.clean_all()`. This also provides a layer of caching, so that it doesn't have to re-clean the input on subsequent calls.

My thought was to provide something similar in here, so that you could have consuming code along the lines of:

```python
scanner = MyScanner("/some/path")
if scanner.issues:
    # Handle issues
```
As @bosatsu mentioned, this may not end up staying here. I'm pretty iffy on whether this is a desirable usage pattern. And as these classes start getting implemented, a "proper" usage pattern should start to shake out. Any input and opinions are also very welcome here.

tartufo/scanner.py
```diff
@@ -90,35 +91,269 @@ def file_path(self) -> Optional[str]:
...
+        if self._rules_regexes is None:
+            self._rules_regexes = config.configure_regexes(
+                self.options.default_regexes,
+                self.options.rules,
+                self.options.git_rules_repo,
+                self.options.git_rules_files,
+            )
+        return self._rules_regexes
+
+    @lru_cache()
```
sjacoby-godaddy, line 172: does lru_cache help much here?

tartufo/scanner.py
```diff
@@ -90,35 +91,269 @@ def file_path(self) -> Optional[str]:
...
+            found_strings = pattern.findall(chunk.contents)
+            for match in found_strings:
+                # Filter out any explicitly "allowed" match signatures
+                if not self.signature_is_excluded(match, chunk.file_path):
+                    issue = Issue(IssueType.RegEx, match)
+                    issue.issue_detail = key
+                    issues.append(issue)
+        return issues
+
+    @abc.abstractproperty
```
sjacoby-godaddy: note: this appears to be depreciated as of version 3.3

sjacoby-godaddy (approved):
Looks good, just a couple non-blocking comments

tartufo/scanner.py
bosatsu: Ahh, I missed the `re.compile`.  I was more thinking of completing the scan for all valid regex's and then logging ones the fail, or notify the user in some other way, but the more I think about it, that would be a new feature, so this is fine as is.

tartufo/scanner.py
tarkatronic: Huh. So it is. Good catch!

tartufo/scanner.py
tarkatronic: I don't have any metrics to back it up, but I'm pretty sure it will.

Imagine a repository with thousands of commits, and only 10 files. This function will get called with the same set of arguments hundreds or even thousands of times, and have to re-check the entire list of regexes every time. This should be a nice shortcut for that.

beardedeagle:
will review this weekend

tests/test_git_repo_scanner.py
```diff
@@ -0,0 +1,225 @@
...
+    def test_repo_is_loaded_on_init(self, mock_repo: mock.MagicMock):
+        scanner.GitRepoScanner(self.options, ".")
+        mock_repo.assert_called_once_with(".")
+
+    @mock.patch("git.Repo")
+    def test_load_repo_loads_new_repo(self, mock_repo: mock.MagicMock):
+        test_scanner = scanner.GitRepoScanner(self.options, ".")
+        test_scanner.load_repo("../tartufo")
+        mock_repo.assert_has_calls((mock.call("."), mock.call("../tartufo")))
+
```
bosatsu, line 29: Just curious on the design for the `GitRepoScanner`, looks like the intention is to insatiate with a git repo but also have the ability to load additional repos?  The first option feels like you would have a scanner object per repo, but the second feels like an object that is a generic scanner that you can feed repos to.  Not necessarily disagreeing with the approach, but was just wondering why.  And I haven't looked to see how this gets wired in, so also wondering if that has something to do with it.

bosatsu (commented):
Finished review, just a last comment/question on design choice.  The new classes are definitely more readable, with the exception of some of git functions, which you already commented.   Also the new unit tests really help with understanding what the code is doing.

tests/test_git_repo_scanner.py
tarkatronic: Similar to the `issues` conversation above, I think this is something that's going to shake itself out as it gets wired in. It may be slightly over-engineered at the moment, and I will work to fix that part. I was trying to abstract out a little bit to allow for solving many use cases in the future, such as:

1. Load a previously cloned local repo
1. Clone a remote repo into a temporary directory and then load that repo
1. Load a bare repo
1. Scanning an entire GitHub org, which would require cloning and loading each repo individually. Although as I've thought further on this, it's likely best to instantiate a new repo scanner per repo.

I would love more discussion on all of this! I'm only one person, so can only present my own views and thoughts. Those conversations can happen here in tickets, or on the GoDaddy-OSS Slack in #tartufo!

bosatsu: 👍  Will follow up on this on the #tartufo channel in GoDaddy-OSS Slack.

bosatsu (approved):

PR #69, opened 2020-09-08
PosixPath type is not JSON serializable

sjacoby-godaddy:
If the cli is used in json mode it consistently givres the following error:
TypeError: Object of type 'PosixPath' is not JSON serializable 

This minor output change resolves that issue


tarkatronic:
This is great, thank you for adding this fix! I am a little bit concerned though that this wasn't caught by our current tests. It looks like that function is covered right now, so that theoretically should have popped up by now: https://codecov.io/gh/godaddy/tartufo/src/master/tartufo/util.py#L31...40

It looks like the tests currently only pass strings for paths though. Could I get you to tweak those tests (`tests.test_util.OutputTests`) to make sure they are using `Path` objects? I want to be sure the tests fail beforehand and provide for a nice regression test.

tarkatronic (approved):
Love it! :shipit:

bosatsu (approved):

PR #70, opened 2020-09-09
Implement new scanners

### LINKED ISSUE #55: tests.test_scanner.ScannerTests.test_return_correct_commit_hash scans actual git history
## 🐛 Bug Report

The test currently looks like this:

```python
    def test_return_correct_commit_hash(self):
        """FIXME: Split this test out into multiple smaller tests w/o real clone
        FIXME: Also, this test will continue to grow slower the more times we commit

        Necessary:
            * Make sure all commits are checked (done)
            * Make sure all branches are checked
            * Make sure `diff_worker` flags bad diffs
            * Make sure all bad diffs are returned
        """
        # Start at commit d15627104d07846ac2914a976e8e347a663bbd9b, which
        # is immediately followed by a secret inserting commit:
        # https://github.com/dxa4481/truffleHog/commit/9ed54617547cfca783e0f81f8dc5c927e3d1e345
        since_commit = "d15627104d07846ac2914a976e8e347a663bbd9b"
        commit_w_secret = "9ed54617547cfca783e0f81f8dc5c927e3d1e345"
        xcheck_commit_w_scrt_comment = "OH no a secret"

        tmp_stdout = six.StringIO()
        bak_stdout = sys.stdout

        # Redirect STDOUT, run scan and re-establish STDOUT
        sys.stdout = tmp_stdout
        try:
            # We have to clone tartufo mostly because TravisCI only does a shallow clone
            repo_path = util.clone_git_repo("https://github.com/godaddy/tartufo.git")
            try:
                scanner.find_strings(
                    str(repo_path),
                    since_commit=since_commit,
                    print_json=True,
                    suppress_output=False,
                )
            finally:
                shutil.rmtree(repo_path)
        finally:
            sys.stdout = bak_stdout

        json_result_list = tmp_stdout.getvalue().split("\n")
        results = [json.loads(r) for r in json_result_list if bool(r.strip())]
        filtered_results = [
            result for result in results if result["commit_hash"] == commit_w_secret
        ]
        self.assertEqual(1, len(filtered_results))
        self.assertEqual(commit_w_secret, filtered_results[0]["commit_hash"])
        # Additionally, we cross-validate the commit comment matches the expected comment
        self.assertEqual(
            xcheck_commit_w_scrt_comment, filtered_results[0]["commit"].strip()
        )
```

No mocks, reading from the file system, checking the actual git history. The biggest problem with this is that, as the history of this project grows, this test continues to take longer.

This needs to be fixed, which may include some actual code cleanup, as the current code is not the most testable.

tarkatronic:
This is a followup to #67, and wires in all the new code, while removing the old code.

Coincidentally, in doing so, it fixes #55! 🎉 
This is also a major step toward implementing #39, #23, and #12.
So, no big deal.

Also, the `cli` module now has barely any logic, and it's mostly contained where it belongs, so this code is far more (re)usable as a library now.


c-jcarlsen-godaddy (approved):

tartufo/cli.py
```diff
@@ -134,56 +134,29 @@ def main(ctx: click.Context, **kwargs: config.OptionTypes) -> None:
...
-            options=options,
-            custom_regexes=rules_regexes,
-            path_inclusions=path_inclusions,
-            path_exclusions=path_exclusions,
-        )
-    else:
-        if options.git_url:
-            repo_path = util.clone_git_repo(options.git_url)
-            remove_repo = True
+    repo_scanner: scanner.ScannerBase
```
sjacoby-godaddy, line 144: pardon my lack of knowledge, should this be an = instead of a : ?

sjacoby-godaddy (approved):
looks good, nice to see the code base compress

tartufo/cli.py
tarkatronic: Good callout, but in this case no. Rather than a variable assignment, this is a [PEP 526](https://www.python.org/dev/peps/pep-0526/) variable annotation. This is just a way of telling all of the tooling what type this variable is. A lot of times this can be inferred, but in this case `mypy` will actually throw a typing conflict if this isn't declared here, because below we assign both `repo_scanner = GitPreCommitScanner(...)` and `repo_scanner = GitRepoScanner(...)`, and it can't figure out what type the variable is actually supposed to be.

PR #71, opened 2020-09-10
updating docs to run current version of tartufo

sjacoby-godaddy:


docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,94 @@ End-to-End Example
...
+      # use a virtual environment!
+      python3.7 -m venv .venv
+      . .venv/bin/activate
+      
+
+   Next: [Install Poetry]  
+
+   Finally: Run ``poetry install``
+
+
```
tarkatronic: This does need to be effectively documented, but it doesn't really belong in this document; this should be moved to the Installation doc instead. The Contributing doc should also probably be updated to point to the Installation doc.

Also, you should not be explicitly creating/using a venv when using Poetry. Instead, you really should let Poetry manage the venv for you. 😄 

So like the Contributing doc in this branch currently shows, you just need to make sure Poetry itself is installed and then run `poetry install`.

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,94 @@ End-to-End Example
...
       GITHUBPROJECT="yourproject"
-      GITHUBREPO="yourrepo.git"
+      GITHUBREPO="myrepo.git"
       GITHUBADDRESS="github.com"
-      tartufo git@${GITHUBADDRESS}:${GITHUBPROJECT}/${GITHUBREPO}
-      # this will spit out a bunch of stuff and end with something like: Results have been saved in /tmp/tmp_jm3hyne
-      TARTUFOSCANPATH="/tmp/tmp_jm3hyne"
-      # this is the number of individual bad password hits tartufo has found:
-      ls ${TARTUFOSCANPATH} | wc -l
+      git clone --mirror git@${GITHUBADDRESS}:${GITHUBPROJECT}/${GITHUBREPO}
```
tarkatronic, line 32: Using `--mirror` creates a "bare" repo, which is [not currently supported](https://github.com/godaddy/tartufo/issues/12).

Instead, we should encourage users to either scan with `tartufo --repo-path <path-to-repo>` or `tartufo git@${GITHUBADDRESS}:${GITHUBPROJECT}/${GITHUBREPO}`. The former will use an already cloned repository, and will explicitly fetch all refs (which I'm assuming is the intent behind using `--mirror`), unless `--branch` is specified. The latter will clone the remote repo for the user and then act the same as though `--repo-path` were specified.

In other words, an explicit clone solely for the purpose of a `tartufo` scan is never necessary. 😄 

Also keep in mind that this usage is going to change before 2.0! #39 is next up in line for implementation, and will split the scans into different "sub-commands". So the commands will be a little bit different. There will be, for example:

* `tartufo scan-local-repo <path-to-repo>`
* `tartufo scan-remote-repo <git-url>`

The exact names of the sub-commands can certainly change, but that should at least get the idea across.

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,94 @@ End-to-End Example
...
-      echo ${#SORTED_UNIQUE_ENTRIES[@]};
-      printf "%s\n" "${SORTED_UNIQUE_ENTRIES[@]}" > badpwz.txt
-      # remove leading/trailing quotes and trailing commas
-      sed -i -e 's/\,$//g' -e 's/\"$//g' -e 's/^\"//g' badpwz.txt
-      
-   Now you have a "bad password" file! Take a look through it, see if anything is wrong, these values will be replaced in your code history
+      # Run Tartufo on your repo and create a list of high entropy items to remove:
+      poetry run tartufo --regex --json --cleanup ${GITHUBREPO} | jq -r '.found_issues[].matched_string' | sort -u > remove.txt
 
+   Now you have a "bad password" file! Take a look through it, see if anything is wrong. This file will be used by ``BFG`` and replace these flagged "bad password" entries with ``***REMOVED***``. It is important that you read through this file to make sure there are not exceptions that you want to remove and exclude with tartufo!
```
tarkatronic: It would be helpful to link to the "Limiting Scans by Signature" and/or "Limiting Scans by Path" in the features doc here. That said, I looked at 2-3 other pages before I found where that was documented, and I wrote the freaking thing! So... I'm thinking the docs need a bit more of a wholesale update so that they flow better. But that's far beyond the scope of just fixing up this doc...

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,94 @@ End-to-End Example
...
 
-      # Cleanup with
+      # Cleanup with BFG
       wget https://repo1.maven.org/maven2/com/madgag/bfg/1.13.0/bfg-1.13.0.jar
-      git clone --mirror git@${GITHUBADDRESS}:${GITHUBPROJECT}/${GITHUBREPO}
       # Make a backup
       cp -r ${GITHUBREPO} backup_${GITHUBREPO}
-      java -jar bfg-1.13.0.jar --replace-text badpwz.txt ${GITHUBREPO}
+
+      do java -jar bfg-1.13.0.jar --replace-text remove.txt ${GITHUBREPO}
```
tarkatronic: Why `do` here?

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,94 @@ End-to-End Example
...
-      echo ${#PEWPEWPEWARR[@]};
-      SORTED_UNIQUE_ENTRIES=($(echo "${PEWPEWPEWARR[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))
-      echo ${#SORTED_UNIQUE_ENTRIES[@]};
-      printf "%s\n" "${SORTED_UNIQUE_ENTRIES[@]}" > badpwz.txt
-      # remove leading/trailing quotes and trailing commas
-      sed -i -e 's/\,$//g' -e 's/\"$//g' -e 's/^\"//g' badpwz.txt
-      
-   Now you have a "bad password" file! Take a look through it, see if anything is wrong, these values will be replaced in your code history
+      # Run Tartufo on your repo and create a list of high entropy items to remove:
+      poetry run tartufo --regex --json --cleanup ${GITHUBREPO} | jq -r '.found_issues[].matched_string' | sort -u > remove.txt
```
tarkatronic: I don't think that we should be instructing users to use `poetry` here, since that applies only to the development version of `tartufo`, and in fact is not even necessary for that. If the user activates the poetry venv (via `poetry shell`), they can just run `tartufo` directly. But again, we shouldn't be specifying that here. If anywhere, that information would go in the installation and/or contributing docs.

tarkatronic: Also, why specify `--regex` but not `--entropy`?

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,94 @@ End-to-End Example
...
+
+
+#. Proceed with cleanup/audit
+
+   Now you have removed the low hanging fruit, it's time to look at the tough stuff
+
+   .. code-block:: console
+
+      #run tartufo again to check for icky code straglers in active code
+      leftovers=`poetry run tartufo --regex --no-cleanup ${GITHUBREPO}`
```
tarkatronic: Same comments here as with the above invocation of `tartufo`

tarkatronic (changes requested):
Overall, I like the updates to the doc. I think it's becoming much more useful. Just a few changes, mostly around the talk of `poetry`. 👍

docs/source/examplecleanup.rst
sjacoby-godaddy: for options specified ref: that one link

docs/source/examplecleanup.rst
sjacoby-godaddy: As the title indicates:
#. Setup poetry if you want to use the most recent build

As this will get merged to master (not stable) and there are still changes in the pipe, it makes sense to have a doc for folks who want to use the codebase in it's current form; ie via poetry. I can add the poetry shell section. 

As for options specified ref: that one link

docs/source/examplecleanup.rst
sjacoby-godaddy: As the title indicates we're working with a non-complete build, leaving those references out seems likely to leaving things to chance for the end user. If we need better install poetry info, the links provided can be augmented or changed to another page (once it exists)

docs/source/examplecleanup.rst
sjacoby-godaddy: neglected to push up the most recent fix, good catch!

docs/source/examplecleanup.rst
tarkatronic: The thing is, this is only a temporary state until 2.0 is released, which is planned ideally for some time this month. This document should be written with the assumption that 2.0 has already been released, so that it doesn't have to be changed again once we release. Note that this document ONLY displayed in the "latest" version of the docs, not the "stable" version, which is what shows to users by default. If you go to https://tartufo.readthedocs.io/ right now, you will not see this document, because it has never been added to the `1.x` branch, which is where `stable` tracks, until 2.0 has been released.

Also, again, this document should only cover *usage*, with installation covered in the Installation doc. Otherwise we have duplicated documentation leading to potentially outdated documentation and even possible user confusion.

docs/source/examplecleanup.rst
sjacoby-godaddy: > Also, why specify `--regex` but not `--entropy`?

Should be enabled by default, no?

  --entropy / --no-entropy        Enable entropy checks. [default: True]
  --regex / --no-regex            Enable high signal regexes checks. [default:
                                  False]

tarkatronic: Yeah I suppose there is that. 😄 That works. I just always tend toward explicit. I guess not necessary here though :)

docs/source/examplecleanup.rst
```diff
@@ -98,19 +130,27 @@ An End-to-End example walkthrough of a tartufo scan and the process of purging t
...
       tartufo git@${GITHUBADDRESS}:${GITHUBPROJECT}/${NEWGITHUBREPO}
-      # should have very little (if any) output. check the newly outputed results
+      # should have very little (if any) output. check the newly outputed results in the given tmp folder
       ls /tmp/tmp_4i4c978 | wc -l
+
+
+Additional Resources
+------------------
+
+[Install Poetry]: https://python-poetry.org/docs/#installation
```
tarkatronic: Looks like this line is no longer needed. 😄

docs/source/examplecleanup.rst
```diff
@@ -98,19 +130,27 @@ An End-to-End example walkthrough of a tartufo scan and the process of purging t
...
+
+[Poetry]: https://python-poetry.org/
```
tarkatronic: It seems that something is not quite right with this link. If you look at the rendered file, line 12 should be getting a link in it and this line should actually disappear.

........and I just realized why. This is a `.rst` file, and that is `.md` link syntax. Take a look at docs/source/index.rst lines 27 and 84 for an example of a named link in ReStructuredText. Also, once this is fixed, the `Additional Resources` header should go away since this will disappear. 😄

docs/source/examplecleanup.rst
tarkatronic: Thoughts on this?

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,87 @@ End-to-End Example
...
+#. OPTIONAL Development only: Setup poetry if you want to use the most recent non-released build from github (may not be stable)
+
+   This project uses [Poetry] to manage its dependencies and do a lot of the heavy lifting. So you'll need to clone the tartufo repo and setup poetry!
+
+
+   .. code-block:: console
+
+      git clone git@github.com:godaddy/tartufo.git
+      
+   Development Use Only Poetry Setup: https://tartufo.readthedocs.io/en/master/CONTRIBUTING.html#setting-up-a-development-environment
```
tarkatronic: Instead of an absolute link to the contributing doc, you can use a Sphinx ref here. I believe what you will want here is

```rst
:doc:`CONTRIBUTING#setting-up-a-development-environment`
```
This makes sure the link works in all environments, across all build hosts, and continues working if the doc URL ever changes.

tarkatronic (changes requested):
Looking much better. Just a few more fixes and it'll be great!

docs/source/examplecleanup.rst
sjacoby-godaddy: yes, that is way better, thank you!

sjacoby-godaddy: oh, but will that reference master?

docs/source/examplecleanup.rst
sjacoby-godaddy: I think "not necessary" may be accurate, but it is "useful"
It's faster to process a large repo locally, it lets you re-run things on the local without pushing them into github until they are just the way you want them (used a few times in this solution), and it allows you to more effectively split up bfg runs (also provided in this solution)

docs/source/examplecleanup.rst
sjacoby-godaddy: Agreed, adding

docs/source/examplecleanup.rst
tarkatronic: Using that syntax will automatically reference whatever version is being built/viewed. So if you're viewing the `master` version of the docs, it will point there, if you're viewing the `latest` version of the docs it will point there, etc.

And once we switch 2.0 over to be our `stable` version, the same will apply there as well. :)

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,106 @@ End-to-End Example
...
+#. OPTIONAL Development only: Setup poetry if you want to use the most recent non-released build from github (may not be stable)
+
+   This project uses `Poetry`_ to manage its dependencies and do a lot of the heavy lifting. So you'll need to clone the tartufo repo and setup poetry!
+
+
+   .. code-block:: console
+
+      git clone git@github.com:godaddy/tartufo.git
+      
+   Development Use Only Poetry Setup: :doc:`CONTRIBUTING` setting-up-a-development-environment
```
tarkatronic: I think this should work to link directly to the appropriate anchor, and also give nice link text.
(This suggestion is no longer good or relevant.)

tarkatronic: I'm liking this section a whole lot more now. This is great!

Looking at it again with fresh eyes has got me wondering: Since it's optional, should it be a part of the numbered bullet points, or would it perhaps serve better (and look better!) as a "note" section? I think it would be interesting to at least try and see what it looks like. It should be very similar to the "Warning" here, but using the light blue colors and say "Note": https://tartufo.readthedocs.io/en/stable/usage.html#python-pre-commit-hook

And theoretically all you have to do to accomplish that is put a `.. note::` line before this, and remove the `#.` so it's no longer a bullet point.

Worth a look, you think?

docs/source/examplecleanup.rst
tarkatronic: Cool, agreed; "mirror" clones have all the info needed for tartufo to work, even if they are "bare" repos under the covers.

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,106 @@ End-to-End Example
...
-      # this is the number of individual bad password hits tartufo has found:
-      ls ${TARTUFOSCANPATH} | wc -l
+      git clone --mirror git@${GITHUBADDRESS}:${GITHUBPROJECT}/${GITHUBREPO}
 
 
-#. Take results and create a "bad password" file. 
+#. Use ``tartufo`` to scan your repository and find any secrets in its history!
 
-   This file will be used by ``BFG`` and replace these flagged "bad password" entries with ``***REMOVED***``. It is important that you read through this file to make sure there are not exceptions that you want to remove and exclude with tartufo!
+   Use what you've learned to scan your repo!
```
tarkatronic: I'm still confused by this line. So far in these steps the user has basically only cloned their repo -- so what is it that they have learned at this point?

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,106 @@ End-to-End Example
...
-      echo ${#SORTED_UNIQUE_ENTRIES[@]};
-      printf "%s\n" "${SORTED_UNIQUE_ENTRIES[@]}" > badpwz.txt
-      # remove leading/trailing quotes and trailing commas
-      sed -i -e 's/\,$//g' -e 's/\"$//g' -e 's/^\"//g' badpwz.txt
-      
-   Now you have a "bad password" file! Take a look through it, see if anything is wrong, these values will be replaced in your code history
+      # Run Tartufo on your repo and create a list of high entropy items to remove:
+      tartufo --regex --json --cleanup ${GITHUBREPO} | jq -r '.found_issues[].matched_string' | sort -u > remove.txt
+
+   Now you have a "bad password" file! Take a look through it, see if anything is wrong. This file will be used by ``BFG`` and replace these flagged "bad password" entries with ``***REMOVED***``. It is important that you read through this file to make sure there are not exceptions that you want to remove and exclude with tartufo! :doc:`features`
```
tarkatronic: Maybe say something like... "Read more about configuring exclusions here:"?

docs/source/examplecleanup.rst
```diff
@@ -6,55 +6,106 @@ End-to-End Example
...
+      cat ${tmppath}/* | jq -r '.signature' | sort -u > allsignatures.txt 
+      sed -i -e 's/$/\",/g' -e 's/^/  \"/g' allsignatures.txt
+      linestr=`grep -n 'exclude-signatures = \[' pyproject.toml`
+      line=`echo $linestr | cut -d ":" -f 1`
+      line=$(($line+1))
+      { head -n $(($line-1)) pyproject.toml; cat allsignatures.txt; tail -n +$line pyproject.toml; } > pyproject.toml_new
+      mv pyproject.toml pyproject.toml_bak
+      mv pyproject.toml_new pyproject.toml
+      # one final run to make sure your signatures are all set
+      poetry run tartufo ${gitrepo} --regex --no-cleanup
```
tarkatronic, line 100: I think this underscores the need for a new command like `tartufo creation-exclusion-list` once #78 is done and merged. 😄

docs/source/examplecleanup.rst
```diff
@@ -84,11 +135,11 @@ An End-to-End example walkthrough of a tartufo scan and the process of purging t
        ! [remote rejected] refs/pull/7/head -> refs/pull/7/head (deny updating a hidden ref)
        ! [remote rejected] refs/pull/8/head -> refs/pull/8/head (deny updating a hidden ref)
        ! [remote rejected] refs/pull/9/head -> refs/pull/9/head (deny updating a hidden ref)
-      error: failed to push some refs to 'git@${GITHUBADDRESS}:yourproject/yourrepo.git'
+      error: failed to push some refs to 'git@GITHUBADDRESS:yourproject/yourrepo.git'
       (.venv) you@LTDV-you:~/tartufo/yourrepo.git$
 
 
-   If you get the above error; It might actually be ok, re-run tartufo. Only if there are results that are not clean continue:
+   If you get the above error; It might actually be ok, re-run tartufo from master. Only if there are results that are not clean continue. Please note, this solution will remove PR history (but not commit history):
```
tarkatronic: Minor pedantic grammar update 😄 
```suggestion
   If you get the above error, it might actually be ok; just re-run tartufo from master. Only continue if there are results that are not clean. Please note, this solution will remove PR history (but not commit history):
```

tarkatronic (changes requested):
Just a few more things. I really want to be sure we get the docs "right" to be as helpful and educational as possible to our users. Apologies if I'm being a bit pedantic here.

Also, I would encourage testing all these doc changes locally with `tox -e docs` and then opening up `docs/build/html/index.html`, to make sure things are looking right.

docs/source/examplecleanup.rst
sjacoby-godaddy: the section is called "would you like to know more" in my opinion this is the example you would goto after you did the pre-requisite reading. if that's confusing I can take it out

docs/source/examplecleanup.rst
sjacoby-godaddy: adding `.. note::` seems to conflict with the code-block section just below

docs/source/examplecleanup.rst
sjacoby-godaddy: 👍  adding

docs/source/examplecleanup.rst
sjacoby-godaddy: yeah, that would be pretty sweet!

docs/source/examplecleanup.rst
sjacoby-godaddy: 👍  modifying

docs/source/examplecleanup.rst
tarkatronic: I think it might be -- just because, you can't control where a user is coming from. I would personally click on "Would you like to know more?" just out of sheer curiosity, and then stick around because it claims to be an "end-to-end example", which I would take to mean "all-inclusive".

docs/source/examplecleanup.rst
tarkatronic: Looks great here when I render it locally! Aside from the "Contributing" link being still a bit off. Or rather, the text after it looking out of place.
Screen Shot 2020-09-14 at 5 16 46 PM

tarkatronic: Also, I found a solution to the link problem, here: https://stackoverflow.com/a/19543591/1971587

It's not ideal, but I realized it's because the `CONTRIBUTING` is a `.md` file still. And that's preventing some of the built-in mechanisms from working. That should probably get converted at some point. Regardless, this works for now:

```suggestion
   Development Use Only Poetry Setup: `Setting up a development environment <CONTRIBUTING.html#setting-up-a-development-environment>`_
```

docs/source/CONTRIBUTING.md
```diff
@@ -107,10 +107,11 @@ lifting. This includes managing development environments! If you are not
 familiar with this tool, we highly recommend checking out [their docs][poetry docs]
 to get used to the basic usage.
 
-Now, setting up a development environment is super simple!
+Now, setting up a development environment is super simple! Additional info if you run into trouble: [Poetry Environments]
 
 Step 1: [Install Poetry]  
 Step 2: Run ``poetry install``
+Step 3: Optionally Run ``poetry shell``
```
tarkatronic: Super hard to tell there is anything wrong here. I only noticed when browsing locally, and because I have visible whitespace in my IDE. In Markdown, you have to put two blank spaces at the end of a line, to force a line break. It's a PITA, I know.
```suggestion
Step 2: Run ``poetry install``  
Step 3: Optionally Run ``poetry shell``
```

sjacoby-godaddy: 👍  nice catch, I missed that

docs/source/examplecleanup.rst
sjacoby-godaddy: updated

docs/source/examplecleanup.rst
sjacoby-godaddy: ah-ha! thanks for the tip!

tarkatronic (approved):

PR #77, opened 2020-09-11
Testing out a CodeQL workflow

tarkatronic:
This is a new GitHub feature for automagic static analysis of repos. Ideally it should help with ensuring secure code. I'm curious to see how this turns out! 🤞

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,46 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master, v1.x]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master, v1.x]
+  schedule:
+    - cron: '0 6 * * 3'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Override automatic language detection by changing the below list
+        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
+        language: ['python']
+        # Learn more...
+        # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,46 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master, v1.x]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master, v1.x]
+  schedule:
+    - cron: '0 6 * * 3'
```
sjacoby-godaddy, line 10: not important, just curious why wed at 6am was selected?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,46 @@
...
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
```
sjacoby-godaddy, line 43: I bet there's a good reason to specify the languages this way vs just putting the equivalent of line 22 in here which is hardcoded anyway, I am not sure what that reason is, fast-fail maybe?

sjacoby-godaddy (approved):
Looks good, follows pattern established in https://github.com/github/codeql-action Usage section pretty close, just a couple questions, non-blocking

.github/workflows/codeql-analysis.yml
tarkatronic: No idea; this was just the default. 😄

.github/workflows/codeql-analysis.yml
tarkatronic: Oh huh yeah I'm not sure. I think this would be more useful for multi-language repos like `asherah`. But shouldn't harm anything here. If anything, I'd rather only configure in one spot.

Issue #75, opened 2020-09-11
The current documentation is fragmented, scattered, and confusing


tarkatronic:
## 📃 Summary

I wrote most of the documentation myself, but even I don't know where most of the bits are. It's confusing and not entirely helpful for the most part.

## Expected documentation

I think it would be helpful to take a holistic view at revamping the docs to be more user friendly. With so much changing in v2.0, I think this is the perfect opportunity to work on this.

Issue #68, opened 2020-09-08
The code itself is poorly and inconsistently documented


tarkatronic:
## 📃 Summary

Most of the code, including code currently in PR (#67), does not have docstrings. This makes the code overall harder to understand, and makes for a steeper developer onboarding curve.

## Expected documentation

All modules, all functions, all classes, and all methods should have docstrings describing them using conventions specified in [PEP 257](https://www.python.org/dev/peps/pep-0257/), and [Sphinx Annotations](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-python-domain).

Ideally this will be checked and enforced via a `pylint` plugin, if such a thing exists. I'm pretty sure it will, since one does exist for `flake8`.

PR #78, opened 2020-09-14
Split the functionality into sub-commands

### LINKED ISSUE #39: Split functionality into sub-commands?

tarkatronic:
Fixes #39 

Now instead of the confusing options and parsing of the options for the different execution paths, they're now all distinctly separated. So now you have:

* `tartufo pre-commit`
* `tartufo scan-local-repo <path>`
* `tartufo scan-remote-repo <url>`

Between this and the other recent work this means that we will easily be able to add more execution modes in the future, such as `scan-local-folder` and `scan-github-org`.

I debated about the repetition of `scan-` at the beginning of the commands, but I think it's the right choice. I'm open to discussion around this though. The reasoning is I would like to be able, in the future, to add other commands like `tartufo create-exclusion-list` to generate a list of signatures to ignore, and `tartufo check-config`, for example.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
   - id: tartufo
     name: tartufo
     entry: tartufo
-    args: [--pre-commit, --cleanup]
+    args: [--cleanup, pre-commit]
     language: python
     pass_filenames: false
   - id: pylint
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
 - id: tartufo
   name: Tartufo
   description: '`tartufo` is a tool for scanning git repositories for secrets/passwords/high-entropy data'
-  entry: tartufo --pre-commit
+  entry: tartufo
   language: python
-  args: [--cleanup]
+  args: [--cleanup, pre-commit]
   pass_filenames: false
--- a/tartufo/commands/scan_local_repo.py
+++ b/tartufo/commands/scan_local_repo.py
@@ -11,8 +11,8 @@ from tartufo.scanner import GitRepoScanner, Issue
 @click.option(
     "--max-depth",
     default=1000000,
-    help="The max commit depth to go back when searching for secrets."
-    " [default: 1000000]",
+    show_default=True,
+    help="The max commit depth to go back when searching for secrets.",
 )
 @click.option("--branch", help="Specify a branch name to scan only that branch.")
 @click.argument(
--- a/tartufo/commands/scan_local_repo.py
+++ b/tartufo/commands/scan_local_repo.py
@@ -1,0 +1,42 @@
+from typing import List, Optional, Tuple
+
+import click
+
+from tartufo import types, util
+from tartufo.scanner import GitRepoScanner, Issue
+
+
+@click.command("scan-local-repo")
+@click.option("--since-commit", help="Only scan from a given commit hash.")
+@click.option(
+    "--max-depth",
+    default=1000000,
+    help="The max commit depth to go back when searching for secrets."
+    " [default: 1000000]",
+)
+@click.option("--branch", help="Specify a branch name to scan only that branch.")
+@click.argument(
+    "repo-path",
+    type=click.Path(exists=True, file_okay=False, resolve_path=True, allow_dash=False),
+)
+@click.pass_obj
+@click.pass_context
+def main(
+    ctx: click.Context,
+    options: types.GlobalOptions,
+    repo_path: click.Path,
+    since_commit: Optional[str],
+    max_depth: int,
+    branch: Optional[str],
+) -> Tuple[str, List[Issue]]:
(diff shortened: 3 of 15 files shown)


.pre-commit-config.yaml
```diff
@@ -4,7 +4,7 @@ repos:
   - id: tartufo
     name: tartufo
     entry: tartufo
-    args: [--pre-commit, --cleanup]
+    args: [--cleanup, pre-commit]
```
sjacoby-godaddy, line 7: no `--` on `pre-commit` ?

.pre-commit-hooks.yaml
```diff
@@ -1,7 +1,7 @@
 - id: tartufo
   name: Tartufo
   description: '`tartufo` is a tool for scanning git repositories for secrets/passwords/high-entropy data'
-  entry: tartufo --pre-commit
+  entry: tartufo
   language: python
-  args: [--cleanup]
+  args: [--cleanup, pre-commit]
```
sjacoby-godaddy, line 6: no -- on pre-commit ?

tartufo/commands/scan_local_repo.py
```diff
@@ -0,0 +1,42 @@
...
+from tartufo.scanner import GitRepoScanner, Issue
+
+
+@click.command("scan-local-repo")
+@click.option("--since-commit", help="Only scan from a given commit hash.")
+@click.option(
+    "--max-depth",
+    default=1000000,
+    help="The max commit depth to go back when searching for secrets."
+    " [default: 1000000]",
```
sjacoby-godaddy: very minor: would it be better to have the help print the default value rather than having a second hard coded entry?

sjacoby-godaddy (approved):
looks good, just a couple questions, nothing blocking

.pre-commit-config.yaml
tarkatronic: Nope! That's the entire point of this change. The "pre-commit" behavior is no longer an option of the main command. It is now an entirely separate sub-command, `tartufo pre-commit`. Just like the behavior triggered by specifying a "--repo-path" previously is now moved to `tartufo scan-local-repo`, and the behavior triggered by specifying a "git-url" (the default invocation) is now moved to `tartufo scan-remote-repo`.

tartufo/commands/scan_local_repo.py
tarkatronic: I agree, that would be ideal. I think that might require a bit of a deeper dive into and customization of `click`, however. The help text is currently entirely auto-generated by `click`. Sounds to me like a good feature request. 😄

tarkatronic: Upon further investigation, this is actually not hard at all! https://click.palletsprojects.com/en/7.x/api/#click.Option

> show_default – controls if the default value should be shown on the help page. Normally, defaults are not shown. If this value is a string, it shows the string instead of the value. This is particularly useful for dynamic options.

I will make it so, for all the things!

sjacoby-godaddy (approved):

mxhenry-godaddy:
Is there a way to preview the ReadTheDocs changes?

tarkatronic:
> Is there a way to preview the ReadTheDocs changes?

You can always view the most recent version of the docs by running `tox -e docs` and then opening up `docs/build/html/index.html` locally. However, this PR does not contain any docs changes. Those will be in a followup PR as part of #68 and #75. Given the size of this PR already and the need to get this implemented and ready quickly, it made the most sense to me to split these up.

djlarsu (approved):
such a good change 👍

sjacoby-godaddy (approved):

PR #81, opened 2020-09-16
Specify working directory for cloning remote repos

### LINKED ISSUE #7: Allow specification of working directory
Support command line switch to override default /tmp working directory location

tarkatronic:
Fixes #7 

@mxhenry-godaddy -- is this what you had in mind for this? Given that this folder is immediately removed when the run is finished, I think the usefulness is a bit dubious. I think the more useful feature will be allowing specification of `--output-dir` via #11.

Let me know!

### DIFF
--- a/tartufo/util.py
+++ b/tartufo/util.py
@@ -10,7 +10,8 @@ import tempfile
 import uuid
 from functools import lru_cache, partial
 from hashlib import blake2s
-from typing import Any, Callable, Dict, Iterable, List, TYPE_CHECKING
+from typing import Any, Callable, Dict, Iterable, List, Optional, TYPE_CHECKING
+from urllib.parse import urlparse
 
 import click
 import git
@@ -57,10 +58,17 @@ def clean_outputs(output_dir: pathlib.Path) -> None:
         shutil.rmtree(output_dir)
 
 
-def clone_git_repo(git_url: str) -> str:
-    project_path = tempfile.mkdtemp()
+def clone_git_repo(
+    git_url: str, target_dir: Optional[pathlib.Path] = None
+) -> pathlib.Path:
+    if not target_dir:
+        project_path = tempfile.mkdtemp()
+    else:
+        repo_name = urlparse(git_url).path.split("/")[-1]
+        project_path = str(target_dir / repo_name)
+
     git.Repo.clone_from(git_url, project_path)
-    return project_path
+    return pathlib.Path(project_path)
 
 
 style_ok = partial(click.style, fg="bright_green")  # pylint: disable=invalid-name
--- a/tartufo/util.py
+++ b/tartufo/util.py
@@ -11,7 +11,6 @@ import uuid
 from functools import lru_cache, partial
 from hashlib import blake2s
 from typing import Any, Callable, Dict, Iterable, List, Optional, TYPE_CHECKING
-from urllib.parse import urlparse
 
 import click
 import git
@@ -64,8 +63,7 @@ def clone_git_repo(
     if not target_dir:
         project_path = tempfile.mkdtemp()
     else:
-        repo_name = urlparse(git_url).path.split("/")[-1]
-        project_path = str(target_dir / repo_name)
+        project_path = str(target_dir)
 
     git.Repo.clone_from(git_url, project_path)
     return pathlib.Path(project_path)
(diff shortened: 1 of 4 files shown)


mxhenry-godaddy:
> Fixes #7
> 
> @mxhenry-godaddy -- is this what you had in mind for this? Given that this folder is immediately removed when the run is finished, I think the usefulness is a bit dubious. I think the more useful feature will be allowing specification of `--output-dir` via #11.
> 
> Let me know!

iirc the incident that spawned the issue was a scan of a very large repo, and thus the working dir was quite large prior to cleanup and the /tmp mount didn't have enough space on the dev's machine. Allowing the working dir to be specified would make it easy to workaround that issue.

tarkatronic:
> > Fixes #7
> > @mxhenry-godaddy -- is this what you had in mind for this? Given that this folder is immediately removed when the run is finished, I think the usefulness is a bit dubious. I think the more useful feature will be allowing specification of `--output-dir` via #11.
> > Let me know!
> 
> iirc the incident that spawned the issue was a scan of a very large repo, and thus the working dir was quite large prior to cleanup and the /tmp mount didn't have enough space on the dev's machine. Allowing the working dir to be specified would make it easy to workaround that issue.

So do you think this would solve that issue?

I have a suspicion that may have also been caused by a sneaky/confusing/dubious bit of behavior in the `tartufo`/`truffleHog` of old which is that if you ran: `tartufo ./my-local-repo`, it would treat `./my-local-repo` as a git URL, rather than a folder name. It would thus create a temp directory and clone that repo into the temp directory. With the separation now of `scan-local-repo` and `scan-remote-repo`, this confusion should no longer occur. So if you have a pre-cloned repo, you won't have to worry about any kind of "working directory"; everything will happen in memory. But, this will help for a repo which is not already cloned.

mxhenry-godaddy:
Sound like we can close #7 as not needed then

tartufo/util.py
```diff
@@ -57,10 +58,17 @@ def clean_outputs(output_dir: pathlib.Path) -> None:
...
-def clone_git_repo(git_url: str) -> str:
-    project_path = tempfile.mkdtemp()
+def clone_git_repo(
+    git_url: str, target_dir: Optional[pathlib.Path] = None
+) -> pathlib.Path:
+    if not target_dir:
+        project_path = tempfile.mkdtemp()
+    else:
+        repo_name = urlparse(git_url).path.split("/")[-1]
+        project_path = str(target_dir / repo_name)
```
dclayton-godaddy: `target_dir` would imply the code would get cloned into that directory. If I said `git clone github.com/myorg/myrepo /tmp/my_target_dir` I would see the code there and not `/tmp/my_target_dir/myrepo`. I would have the caller append repo_name.

tarkatronic: I had that implementation originally and found that it lead to potentially confusing behavior due to the fact that we remove the repo after it's scanned. So let's say I ran:

`tartufo scan-remote-repo --work-dir=~/temp https://github.com/godaddy/tartufo.git`

It would clone the repository (without a containing folder) directly into `~/temp` and scan it there. It would then remove the entire `~/temp` folder, which is NOT what I would expect as an end user.

It felt to me far more reasonable to clone into `~/temp/tartufo.git/` and then remove that `tartufo.git` folder afterward, leaving the target directory in tact, the way it was before hand.

dclayton-godaddy: Correct. You are referring to `work-dir`. I agree. I'm referring to this method. `clone` and `target_dir` I would imagine would operate similar to `git clone` with target dir.

tarkatronic: Ah. So you are proposing to move this `target_dir / repo_name` bit to be moved into the `scan-remote-repo` command then? I think that makes sense. I'll just have to make sure this method gets documented to make that absolutely clear.

sjacoby-godaddy (approved):
nice addition!

tartufo/util.py
tarkatronic: @dclayton-godaddy There ya go, I shifted the path manipulation up to the command itself. Should cover your concerns. Good call on that.

bosatsu (approved):
Didn't test this locally, but code looks good, and tests look much better now.

dclayton-godaddy (approved):

PR #82, opened 2020-09-16
Add a --output-dir/-od option in place of --cleanup

### LINKED ISSUE #11: Optimize handling of temp files (vs --cleanup)

tarkatronic:
Fixes #11 -- @rscottbailey does this work for what you were thinking on this?

Now, there is no `--cleanup` flag. There is only `--output-dir`. If you don't specify one, nothing ever get written to disk. What a revelation!

I also changed the key in the commands JSON output from `issues_path` to `output_dir`, to match. And I took the liberty of appending `.json` to the filenames being output. Because they are JSON files, and file extensions are a good thing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
   - id: tartufo
     name: tartufo
     entry: tartufo
-    args: [--cleanup, pre-commit]
+    args: [pre-commit]
     language: python
     pass_filenames: false
   - id: pylint
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,5 @@
   description: '`tartufo` is a tool for scanning git repositories for secrets/passwords/high-entropy data'
   entry: tartufo
   language: python
-  args: [--cleanup, pre-commit]
+  args: [pre-commit]
   pass_filenames: false
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -2,7 +2,6 @@
 
 import importlib
 import pathlib
-from tempfile import mkdtemp
 from typing import List, Tuple
 
 import click
@@ -94,11 +93,18 @@ class TartufoCLI(click.MultiCommand):
     "you would like.",
 )
 @click.option(
-    "--cleanup/--no-cleanup",
-    is_flag=True,
-    default=False,
-    show_default=True,
-    help="Clean up all temporary result files.",
+    "-od",
+    "--output-dir",
+    type=click.Path(
+        exists=False,
+        file_okay=False,
+        dir_okay=True,
+        writable=True,
+        resolve_path=True,
+        allow_dash=False,
+    ),
+    help="If specified, all issues will be written out as individual JSON files "
+    "in this directory.",
 )
 @click.option(
     "--git-rules-repo",
@@ -151,19 +157,16 @@ def process_issues(
 ):
     repo_path, issues = result
     options = types.GlobalOptions(**kwargs)  # type: ignore
-    if issues:
-        output_dir = pathlib.Path(mkdtemp())
-        util.write_outputs(issues, output_dir)
-    else:
-        output_dir = None  # type: ignore
-    util.echo_issues(issues, options.json, repo_path, output_dir)
-
-    if options.cleanup:
-        util.clean_outputs(output_dir)
-    else:
-        if output_dir and not options.json:
-            click.echo("Results have been saved in {}".format(output_dir))
+    output_dir = None
+    if options.output_dir:
+        output_dir = pathlib.Path(options.output_dir)
 
     if issues:
+        util.echo_issues(issues, options.json, repo_path, output_dir)
+        if output_dir:
+            util.write_outputs(issues, output_dir)
+            if not options.json:
+                click.echo(f"Results have been saved in {output_dir}")
         ctx.exit(1)
+
     ctx.exit(0)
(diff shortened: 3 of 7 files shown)


dclayton-godaddy:
Is the default behavior to write output to stdout?

dclayton-godaddy:
I assume this will be tracked as a 2.0 breaking change.

tarkatronic:
> Is the default behavior to write output to stdout?

The current behavior is:

* Display found issues to `stdout`
* *Always* create a temp directory under `/tmp` using `mkdtemp()`
* *Always* write the issues into files under that temp directory
* If `--cleanup` is specified on the command line, delete that temp directory

This new behavior is:

* Display found issues to `stdout`
* If an output directory was specified, write issues to that directory

> I assume this will be tracked as a 2.0 breaking change.

Yes, #11 is part of the 2.0 milestone, and this is one of the many breaking changes happening in the work toward 2.0.

bosatsu (approved):
Looks good, nothing that stood out to me.

rscottbailey (changes requested):
This looks good but after a little quick usability testing, I think we need a little tweaking... I think the fixes should be quick and easy, but I'm not sure which is better. In short, I'd like to see some low-cost guardrails for stupid human behavior that is likely to occur regularly.

What I did was do some test scans on the CirrusScan repo, which includes some SVG diagrams that happen to have high-entropy content (either as a result of compression, or stored bitmaps, or something else - doesn't really matter). Of course, tartufo hates them (as expected). The output directory option worked as advertised. But...

If you run a scan with `-od`, fix your exclude file (for example), and run the scan again with the same output directory, you end up with the original scan stuff commingled with the new scan stuff. I expect (without having verified) that for any particular finding there's an exact file overwrite, but it is difficult to see what's old (only) and what's new (only) if you have a non-trivial number of findings.

This leads to the argument that to avoid this result, we should insist the output directory either is empty or does not exist (prior to scanning) so when the smoke clears the content unambiguously reflects (only) the last scan. This is basically what I believe the released version does, since it's using a scratch directory. 

Another option would be to create a new subdirectory beneath the output directory, but this is a little non-intuitive, and either the name is weird (UUID, random, etc) or you have to worry about collisions. Yet another option would be to delete files in the output directory that look like they might be findings that weren't found this time, but I think that's just insane for all kinds of reasons.

I believe the easiest thing to implement would be to insist on creating the directory ourselves, because then we only need a single existence check followed by a mkdir.

Sort of following this train of thought, I then ran another scan and specified an output directory that did not exist. What happened is that I churned for quite a bit and then got a page-long stack trace when we attempted to write the first finding file and couldn't because the parent directory was missing. This behavior sucks because I burned a lot of time, got an inscrutable error, and then didn't get ANY of the scan results -- i.e. wasted effort.

The minimally correct thing to do would be to verify the output directory exists before doing anything costly, and then abort quickly with a sane error message if it does not.

Arguably the better thing to do, especially in light of the first discussion, would be to create the output directory when it does not exist. I lean slightly toward creating it on a JIT basis, so that if you run a scan that doesn't find anything, you don't create an empty directory. However, I could see that some users might prefer the certainty of always getting an output directory when requested, even if it sometimes (usually?) ends up empty.

The TL;DR version boils down to:

1. before scan, possibly check if output directory exists (assuming it is specified) - depending on religious persuasion, either abort nicely if it exists or abort nicely if it doesn't exist
2. when writing findings to output directory, possibly before the scan or possibly after the scan, (silently) create the output directory if it does not exist -- obviously unneeded if we did a DNE abort in first step

tarkatronic:
Wow, thanks for the thorough writeup, @rscottbailey!

I was able to duplicate the error behavior you described, and I have a simple 2-line fix for that. I'll push that up as soon as I add a test for it.

As for the conflicting/old scan results... what if I were to do something similar to what I'm doing in #81 and put the results into a subdirectory of the specified directory. In this case though, instead of just calling it (for example) `foo.git`, I could name it something like... `tartufo-scan-results-2020-09-17T11:06:49.410887` -- that last part being the output of `datetime.datetime.now().isoformat()`. That will give us a timestamp with microsecond precision, which should theoretically always be unique.

How does that sound?

rscottbailey:
The timestamp-based subdirectory would be great for me - avoids all the issues I mentioned and it's understandable too.

Thanks!

tarkatronic:
@rscottbailey Okay, I've made the changes you suggested. Another review would be much appreciated!

rscottbailey (approved):
Perfecto - thanks!

PR #101, opened 2020-09-28
Fixes #86: Prefer tartufo.toml over pyproject.toml if both are present

### LINKED ISSUE #86: If a `pyproject.toml` is present, `tartufo.toml` is never checked for configuration
## 🐛 Bug Report

In Python projects, the `pyproject.toml` file is becoming somewhat ubiquitous. It is being used for configuration of all forms of tooling, and even `tartufo` allows users to place their config in this file. However, if users opt to, instead, put their `tartufo` config into a `tartufo.toml` file when they already have a `pyproject.toml` file, this will have no effect.

This is especially problematic once users start using hash-based signatures. These signatures, themselves, will get detected and reported as high entropy strings, meaning that the `tartufo` configuration file should be ignored. In practice, it should generally be perfectly acceptable to ignore the `tartufo.toml`, but ignoring an entire `pyproject.toml`, which could have some legitimate secrets, should be avoided.

## To Reproduce

* Create a `pyproject.toml` file, even an empty one
* Create a `tartufo.toml` file, with a `[tool.tartufo]` section and actual config rules
* Run `tartufo`
* The configuration found in the `tartufo.toml` will not be read nor used.

## Expected Behavior

If a `tartufo.toml` is present, that should likely take precedence over `pyproject.toml`. Also, a `pyproject.toml` with no `[tool.tartufo]` section should not prevent the `tartufo.toml` from being read.

The simplest and most obvious fix might just be to swap the order in which the files are checked: Check for a `tartufo.toml` first; if that doesn't exist, check for a `pyproject.toml`. This will need to be done in both the `config.read_pyproject.toml`, and in the code that checks repositories for their local exclusion rules. (Does this still exist? Did this get removed in the work up to 2.0?)

Rashmi-K-A:
Fixes #86 
- Implemented the solution suggested as part of bug description. 
- Added tests to ensure tartufo.toml takes precedence.

### DIFF
--- a/tartufo/config.py
+++ b/tartufo/config.py
@@ -41,11 +41,11 @@ def read_pyproject_toml(
         if not root_path:
             root_path = "."
         root_path = pathlib.Path(root_path).resolve()
-        config_path = root_path / "pyproject.toml"
+        config_path = root_path / "tartufo.toml"
         if config_path.is_file():
             value = str(config_path)
         else:
-            config_path = root_path / "tartufo.toml"
+            config_path = root_path / "pyproject.toml"
             if config_path.is_file():
                 value = str(config_path)
             else:
--- a/tests/data/multiConfig/pyproject.toml
+++ b/tests/data/multiConfig/pyproject.toml
@@ -1,0 +1,2 @@
+[tool.tartufo]
+regex = false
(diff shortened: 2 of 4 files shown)


tarkatronic (approved):
@Rashmi-K-A This is great! Nice and simple, I love it.

Thank you!

PR #102, opened 2020-09-30
feat(scanner): add support for scanning folders

### LINKED ISSUE #83: Support Scanning Directory

We would like to scan the current directory. From this scan, we can remediate any active secret usage and allow tools like `bfg` to clean git history. We only are about the secrets currently being used.

A CLI command like `tartufo scan .`.

Rumor has it we can use `bfg` to temporarily clean history, then rerun tartufo, but that is painful.

Something like, "To run a scan against a directory, use `tartufo scan [{options}] {directory}`."

dclayton-godaddy:
Add support for scanning folders.


tartufo/commands/scan_folder.py
```diff
@@ -0,0 +1,31 @@
+from typing import List, Optional, Tuple
+
+import click
+
+from tartufo import types, util
+from tartufo.scanner import FolderScanner, Issue
+
+
+@click.command("scan-folder")
+@click.option("--pattern", default="*", help="Glob expression used to filter files.")
```
tarkatronic: This should be more explicit about whether this is an inclusion or exclusion filter. As an end user, I wouldn't know just from this. Perhaps the option name could be updated to be more explicit?

tartufo/commands/scan_folder.py
```diff
@@ -0,0 +1,31 @@
...
+@click.argument(
+    "folder-path",
+    type=click.Path(exists=True, file_okay=False, resolve_path=True, allow_dash=False),
+)
```
tarkatronic, line 16: Side note, partially to myself, I'm starting to think that the name of the final argument to all of these sub-commands should be unified as `target`. The differing names could lead to sneaky bugs, as seen here: https://github.com/godaddy/tartufo/pull/103/files#diff-7a641301879c03cbef06773babd13450R113-R116

tartufo/commands/scan_folder.py
```diff
@@ -0,0 +1,31 @@
...
+@click.pass_obj
+@click.pass_context
+def main(
+    ctx: click.Context,
+    options: types.GlobalOptions,
+    folder_path: str,
+    pattern: Optional[str],
+) -> Tuple[str, List[Issue]]:
+    """Scan a folder."""
+    folder_options = types.FolderOptions(pattern=pattern,)
```
tarkatronic: Since there is only the one additional option on this command, I don't think that warrants a whole new `dataclass`. I'm not really sure where that line is (3 options like `GitOptions`?) but it just feels excessive for a single one. On the other hand, I can see the advantage of sticking with a particular convention like this. So this is more of a question mark than a blocker.

tartufo/commands/scan_folder.py
```diff
@@ -0,0 +1,31 @@
...
+    issues: List[Issue] = []
+    try:
+        scanner = FolderScanner(options, folder_options, folder_path)
+        issues = scanner.scan()
+    except types.TartufoException as exc:
+        util.fail(str(exc), ctx)
+    return str(folder_path), issues
```
tarkatronic: Since the `folder_path` is already of type `str`, this cast should be unnecessary

tartufo/scanner.py
```diff
@@ -180,7 +181,7 @@ def should_scan(self, file_path: str):
         this function when no inclusions or exclusions exist.
 
         :param file_path: The file path to check for inclusion
-        :return: False if the file path is _not_ matched by `self.indluded_paths`
+        :return: False if the file path is _not_ matched by `self.included_paths`
```
tarkatronic, line 243: Whoops, good catch! 👍

tartufo/scanner.py
```diff
@@ -249,7 +250,7 @@ def scan(self) -> List[Issue]:
             if self.global_options.entropy:
                 issues += self.scan_entropy(chunk)
         self._issues = issues
-        return self._issues
+        return issues
```
tarkatronic: I can see the logic of returning the local rather than instance variable, but this was intentionally following the pattern of "set instance variable, return instance variable" similar to all of the properties to above. There was definitely discussion about whether that is a pattern we want to support in this library, specifically with issues (https://github.com/godaddy/tartufo/pull/67#discussion_r483291870).

I do question, though, why is this change necessary and relevant to this new feature?

tartufo/scanner.py
```diff
@@ -476,3 +477,52 @@ def chunks(self):
...
+        """Yield the individual file changes currently staged for commit.
+
+        :rtype: Generator[Chunk, None, None]
+        """
+
+        for blob, file_path in self._iter_folder():
+            if self.should_scan(file_path):
+                yield types.Chunk(blob, file_path)
+
+    def _iter_folder(self):
```
tarkatronic: This needs a return annotation. I think it should also accept an argument of the path to be scanned, following the pattern of the generators on the other scanners.

tartufo/scanner.py
```diff
@@ -476,3 +477,52 @@ def chunks(self):
...
+        folder_path = pathlib.Path(self.folder_path)
+        for file_path in folder_path.glob(self.folder_options.pattern):
```
tarkatronic: By default, with your pattern of `*`, this is only going to return a top-level listing of the directory. I think there are two potential routes to consider here.

1. Use [rglob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob) instead, which will prepend `**/` to the pattern and automatically make this recursive
2. Consider a `--recursive` option, defaulted to `True`, and manually walk the tree.

I think both have their pros and cons. The former is definitely easier to implement, but then requires the user to fully understand glob syntax, which can be tricky for more complex situations. The latter is far more explicit and intuitive, and I think we could even get super fancy with the options doing something like `tartufo scan-folder --recursive --include-ext py --include-ext yml` and similar `--exclude-ext` options. But this is a whole big rabbit hole that would be much more complicated to implement, and most likely slower to execute as well.

On the topic of performance though, there is one other thing to note. There is a note in the [docs for glob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) which states:
> Note Using the “**” pattern in large directory trees may consume an inordinate amount of time.

What that "inordinate amount of time" actually means, and what alternatives in `pathlib` would be, I'm not sure. I suppose for pure raw speed in a manual walking scenario, we would want to use [os.walk](https://docs.python.org/3/library/os.html#os.walk) and yield `Path` objects only as necessary.

tartufo/scanner.py
```diff
@@ -476,3 +477,52 @@ def chunks(self):
...
+            if file_path.is_file():
+                with open(file_path, "rb") as fhd:
+                    data = fhd.read()
+
+                try:
+                    blob = data.decode("utf-8")
+                except UnicodeDecodeError:
+                    # binary file, continue
+                    continue
```
tarkatronic, line 729: I'm wondering if this is really a 100% accurate test for a binary file. Does this work with files in other non-UTF8 characters sets such as CP1252? I think these are probably less common nowadays, but not non-existent. I also wonder about the performance of attempting to decode an entire file, vs just reading a few bytes from the top of the file to determine its type. How would this perform with, say, a several hundred megabyte file?

tartufo/scanner.py
```diff
@@ -476,3 +477,52 @@ def chunks(self):
...
+
+        for blob, file_path in self._iter_folder():
+            if self.should_scan(file_path):
+                yield types.Chunk(blob, file_path)
+
+    def _iter_folder(self):
+        folder_path = pathlib.Path(self.folder_path)
+        for file_path in folder_path.glob(self.folder_options.pattern):
+            if file_path.is_file():
+                with open(file_path, "rb") as fhd:
```
tarkatronic, line 722: Neat trick with `Path` objects: You can actually use `with file_path.open("rb") as hfd:`. Not a blocker, just a handy usability thing.

tartufo/types.py
```diff
@@ -27,6 +27,11 @@ class GitOptions:
     branch: Optional[str]
 
 
+@dataclass
+class FolderOptions:
+    pattern: Optional[str] = "*"
```
tarkatronic: Default values for options should be set in the command option, rather than in these classes; that way, it is fully transparent and obvious to the user.

tarkatronic (changes requested):
This is a great start, thank you for the work on this @dclayton-godaddy! This is a functionality I've definitely wanted to see for some time. And I'm glad to see that the recent codebase changes have made this relatively straightforward to implement!

I've got a few changes I'd like to see in here, as well as some tips and just general food for thought.

I'm looking forward to getting this buttoned up and ready -- looks like version 2.1 may come sooner than I expected!

tartufo/commands/scan_folder.py
dclayton-godaddy: Should I report this as a bug then? https://github.com/godaddy/tartufo/blob/master/tartufo/commands/scan_local_repo.py#L46

tarkatronic: Ah, good catch. I wouldn't call it a bug since it's not breaking anything, but it's also unnecessary. I'm pretty sure that's a leftover artifact from when I was trying to cast those as `click.Path`, while it turns out that `click` actually just passes in an `str`.

tartufo/scanner.py
dclayton-godaddy: Do the git scanners require `--recursive`? Glob by default enables recursion. `*` versus `**/*`. https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

tartufo/scanner.py
dclayton-godaddy: Why would we pass the path in if the path is passed into the constructor and the `_iter_folder` is an internal method to the class? If this needs to be reused internally, cant it be refactored then? Seems more of a personal preference.

tartufo/scanner.py
dclayton-godaddy: nice!

tartufo/types.py
dclayton-godaddy: This enables the CLI and the programmatic interfaces maintain the same defaults.

dclayton-godaddy: I prefer to write all python code as an API and CLI is but one consumer of that API.

tartufo/scanner.py
dclayton-godaddy: Should I create a bug for this as well? https://github.com/godaddy/tartufo/blob/master/tartufo/scanner.py#L342

dclayton-godaddy: IMHO this is bad practice checking the contents of an error message `if printable_diff.startswith("Binary files"):` as the error message should not be contractual.

tartufo/commands/scan_folder.py
dclayton-godaddy: This goes in line with the dozens of bash commands there they reference "PATTERN" as an implicit include. Includes and excludes can be handled through the global options.

tartufo/commands/scan_folder.py
dclayton-godaddy: I think this is more explicit and easier to read. I would stick with named. Can we assume all commands' first argument will be a "target"?

tartufo/commands/scan_folder.py
dclayton-godaddy: If you look at many, if not most, bash commands, the arguments that filter refer to the value as a "PATTERN" as an implicit include glob expression.

tartufo/commands/scan_folder.py
dclayton-godaddy: This is a better convention and allows for additional options in the future. We also want folder explicit options to remain as sub command options and not global options.

tartufo/commands/scan_folder.py
dclayton-godaddy: Also, if you notice the actual `glob` function, with the argument `pattern` makes no mention of `include` or `exclude`. https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

tartufo/scanner.py
jgowdy: I would agree that instance methods don't need to take member fields as parameters.

jgowdy:
What is the state of this PR?  It seems like it didn't get merged with the 2.1/2.2 release that went out.

dclayton-godaddy:
@tarkatronic I rebased this and addressed most of your concerns. Let me know if there is anything else that would block this from being merged. Thanks!

dclayton-godaddy:
Going to open a new PR with some improved changes.

PR #108, opened 2020-10-05
feat(scan): use generator to display live output

dclayton-godaddy:
The current process does not output anything until the entire scan is complete, this enables the scan to immediately display output to allow the user to CTRL-C, add ignores, and rerun.


jgowdy (approved):

c-jcarlsen-godaddy (approved):

tarkatronic:
It took me a while, but I just finally remembered why I had moved away from live/streaming output like this! In the original implementation, it ended up outputting invalid JSON output, because it was printing out a bunch of individual items instead of one big JSON object.

I don't know for sure if that will be the case here, but I want to be sure that is covered. We need to be able to continue to pipe JSON output to `jq`.

dclayton-godaddy:
`json.dumps` does not support a generator. The following line returns a serializable list. https://github.com/godaddy/tartufo/blob/master/tartufo/util.py#L56

jgowdy:
What is the state of this PR?  If we're trying to make one big parsable JSON blob it seems like we could put array syntax like "{ [ " followed by each JSON fragment, followed by ',' and finish with "] }"?

pmevzek-godaddy:
FWIW , there is the notion of "JSON Streaming".  Based of https://en.wikipedia.org/wiki/JSON_streaming `jq` would continue to work in some of the cases (line or concatenated versions)

rbailey-godaddy:
I looked at this and then reimplemented it as PR #227 -- the code is quite similar, but I encountered weird issues when trying to rebase this PR up to current main and it was easier to just follow your lead starting with a clean branch. Please follow the new PR.

PR #111, opened 2020-10-07
Updated docs for v2

### LINKED ISSUE #74: Consider using sphinx-click
## 📃 Summary

Right now, the usage shown both in the `README.md` and the `docs/configuration.rst` is copy-pasted by hand. This is error prone and can easily become outdated. Especially as #39 gets implemented, this situation will only become worse. It'd be great if we could automate this somehow.

[sphinx-click](https://pypi.org/project/sphinx-click/) claims to provide that. So, we should try it out and see if we can make it work for our needs. That would be ideal!

## Expected documentation

We will want automatically updated documentation of the usage of the primary command, as well as all sub-commands.

### LINKED ISSUE #75: The current documentation is fragmented, scattered, and confusing

tarkatronic:
This is, hopefully, the last thing needed in the lead-up to v2.0.0 final!

Fixes #74 
Fixes #75 

This ended up, largely, being a rewrite and cleanup of the documentation. The hope is that the structure overall makes more sense and information is more readily available, easier to find, and automatically up-to-date.

I added 2 Sphinx modules:

* `sphinx-click` -- This automatically extracts the full usage of the `click` command, for the `Usage` document. This way we don't have to remember to copy-paste it into there every time usage changes. **NOTE:** This does not, however, fix that need for the `README.md`. Not sure what to do there yet.
* `sphinxcontrib-spelling` -- A spellchecker! Yay! This also meant setting up a custom word list under `docs/source/spelling_wordlist.txt`, for words that we know are definitely spelled right, but aren't in the default dictionary.

Possibly the easiest way to review these will be with the fully rendered version, so I've activated this branch over on ReadTheDocs (although it is hidden from the version menu). You can find it here: https://tartufo.readthedocs.io/en/v2-docs/


docs/source/examplecleanup.rst
```diff
@@ -56,53 +77,58 @@ An End-to-End example walkthrough of a tartufo scan and the process of purging t
       cp -r ${GITHUBREPO} backup_${GITHUBREPO}
       java -jar bfg-1.13.0.jar --replace-text remove.txt ${GITHUBREPO}
 
-
 #. Uh Oh!
 
-   Occasionally the results will be to big to process all at once. If that happens simply split up the results and loop through them.
+   Occasionally the results will be to big to process all at once. If that
```
mxhenry-godaddy: `will be too big to process`

docs/source/examplecleanup.rst
```diff
@@ -56,53 +77,58 @@ An End-to-End example walkthrough of a tartufo scan and the process of purging t
...
 
-   Now you have removed the low hanging fruit, it's time to look at the tough stuff
+   Now you have removed the low hanging fruit, it's time to look at the tough
+   stuff
 
    .. code-block:: console
 
-      #run tartufo again to check for icky code straglers in active code
-      leftovers=`tartufo --regex --no-cleanup ${GITHUBREPO}`
+      # run tartufo again to check for icky code straglers in active code
```
mxhenry-godaddy: `code stragglers in`

docs/source/examplecleanup.rst
```diff
@@ -111,9 +137,9 @@ An End-to-End example walkthrough of a tartufo scan and the process of purging t
       git push
 
 
-#. Danger Will Robinson, Danger! 
+#. Danger Will Robinson, Danger!
```
mxhenry-godaddy, line 144: Probably want to add a note that anyone that has cloned the repo prior to this should re-clone to prevent reintroducing the bad history

mxhenry-godaddy (approved):
Love the wordsmithing and things are very clear. Just a few nitpicks.

docs/source/examplecleanup.rst
tarkatronic: @mxhenry-godaddy Good point! Do you think this will suffice? https://github.com/godaddy/tartufo/pull/111/files#diff-cd26aa27595802087598d66aa01eac4dR128-R135

Visible here at point number 7: https://tartufo.readthedocs.io/en/v2-docs/examplecleanup.html

mxhenry-godaddy: yep

sjacoby-godaddy (approved):
looks great, ship it!

docs/source/examplecleanup.rst
```diff
@@ -56,53 +77,62 @@ An End-to-End example walkthrough of a tartufo scan and the process of purging t
...
 
-   Now you have removed the low hanging fruit, it's time to look at the tough stuff
+   Now you have removed the low hanging fruit, it's time to look at the tough
+   stuff
 
    .. code-block:: console
 
-      #run tartufo again to check for icky code straglers in active code
-      leftovers=`tartufo --regex --no-cleanup ${GITHUBREPO}`
+      # run tartufo again to check for icky code stragglers in active code
```
jgowdy: This type of idiomatic language might be difficult to parse for non-native English speakers.

rfox-godaddy (approved):

docs/source/examplecleanup.rst
tarkatronic: Good point. Do you have some recommended improved verbiage?

dclayton-godaddy: Maybe `run tartufo again to check for potential secrets in the latest commit`

tarkatronic: I'm not sure that was the *exact* intent so I left it a bit more open-ended:
> run tartufo again to check for any remaining potential secrets

@jgowdy That work for ya?

jgowdy: 👍

PR #116, opened 2020-10-12
feat(rules): Add path specific rules (#88)

### LINKED ISSUE #88: File Extension-based Matching

To detect literal passwords in files like YAML, JSON, JS, etc require different matching.

Examples:

* yaml: `password: mypassword`
* json: `"password": "mypassword"`
* js: `{password: 'mypassword'}`

Add the file extension to regexs file.

```yaml
- message: contains a password or token
  match: '(password|passwd|pwd|token|pass)(["])?[=:]'
  conditions:
    - file: '*.js'
      match: '(password|passwd|pwd|token|pass)([''" \t]*)[=:]([ \t]*)(([''"][^''"])|([\[]))'
    - file: '*.json'
      match: '(password|passwd|pwd|token|pass)([''" \t]*)[=:]([ \t]*)(([''"][^''"])|([\[]))'
    - file: '*.py'
      match: '(password|passwd|pwd|token|pass)([''" \t]*)[=:]([ \t]*)(([''"][^''"])|([\[]))'
```

dclayton-godaddy:
This will add the ability to define rules for files that match a specific pattern. For example, YAML files might need to match without quotes, whereas JS/JSON will need to match with quotes.

Closes #88

### DIFF
--- a/tartufo/config.py
+++ b/tartufo/config.py
@@ -1,3 +1,4 @@
+import collections
 import copy
 import json
 import pathlib
@@ -22,11 +23,11 @@ import toml
 import truffleHogRegexes.regexChecks
 
 from tartufo import types, util
-
+from tartufo.types import Rule
 
 OptionTypes = Union[str, int, bool, None, TextIO, Tuple[TextIO, ...]]
 
-DEFAULT_REGEXES = truffleHogRegexes.regexChecks.regexes
+DEFAULT_REGEXES = util.convert_regexes_to_rules(truffleHogRegexes.regexChecks.regexes)
 
 
 def load_config_from_path(
@@ -151,7 +152,7 @@ def configure_regexes(
     rules_files: Optional[Iterable[TextIO]] = None,
     rules_repo: Optional[str] = None,
     rules_repo_files: Optional[Iterable[str]] = None,
-) -> Dict[str, Pattern]:
+) -> Dict[str, Rule]:
     """Build a set of regular expressions to be used during a regex scan.
 
     :param include_default: Whether to include the built-in set of regexes
@@ -195,22 +196,34 @@ def configure_regexes(
     return rules
 
 
-def load_rules_from_file(rules_file: TextIO) -> Dict[str, Pattern]:
+def load_rules_from_file(rules_file: TextIO) -> Dict[str, Rule]:
     """Load a set of JSON rules from a file and return them as compiled patterns.
 
     :param rules_file: An open file handle containing a JSON dictionary of regexes
     :raises ValueError: If the rules contain invalid JSON
     """
-    regexes = {}
+    rules: Dict[str, Rule] = {}
     try:
         new_rules = json.load(rules_file)
     except json.JSONDecodeError as exc:
         raise ValueError(
             "Error loading rules from file: {}".format(rules_file.name)
         ) from exc
-    for rule in new_rules:
-        regexes[rule] = re.compile(new_rules[rule])
-    return regexes
+    for rule_name in new_rules:
+        rule_definition = new_rules[rule_name]
+        if isinstance(rule_definition, collections.Mapping):
+            path_pattern = rule_definition.get("path_pattern", None)
+            rule = Rule(
+                name=rule_name,
+                pattern=re.compile(rule_definition["pattern"]),
+                path_pattern=re.compile(path_pattern) if path_pattern else None,
+            )
+        else:
+            rule = Rule(
+                name=rule_name, pattern=re.compile(rule_definition), path_pattern=None
+            )
+        rules[rule_name] = rule
+    return rules
 
 
 def compile_path_rules(patterns: Iterable[str]) -> List[Pattern]:
--- a/tartufo/types.py
+++ b/tartufo/types.py
@@ -1,7 +1,7 @@
 # pylint: disable=too-many-instance-attributes
 import enum
 from dataclasses import dataclass, field
-from typing import Any, Dict, Optional, TextIO, Tuple
+from typing import Any, Dict, Optional, TextIO, Tuple, Pattern
 
 
 @dataclass
@@ -37,6 +37,13 @@ class Chunk:
     contents: str
     file_path: str
     metadata: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class Rule:
+    name: Optional[str]
+    pattern: Pattern
+    path_pattern: Optional[Pattern]
 
 
 class TartufoException(Exception):
--- a/tartufo/util.py
+++ b/tartufo/util.py
@@ -9,12 +9,13 @@ import tempfile
 import uuid
 from functools import lru_cache, partial
 from hashlib import blake2s
-from typing import Any, Callable, Dict, Iterable, List, Optional, TYPE_CHECKING
+from typing import Any, Callable, Dict, Iterable, List, Optional, TYPE_CHECKING, Pattern
 
 import click
 import git
 
 from tartufo import types
+from tartufo.types import Rule
 
 if TYPE_CHECKING:
     from tartufo.scanner import Issue  # pylint: disable=cyclic-import
@@ -34,6 +35,15 @@ def del_rw(_func: Callable, name: str, _exc: Exception) -> None:
     """
     os.chmod(name, stat.S_IWRITE)
     os.remove(name)
+
+
+def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
+    rules: Dict[str, Rule] = {}
+    for regex_name in regexes:
+        rules[regex_name] = Rule(
+            name=regex_name, pattern=regexes[regex_name], path_pattern=None
+        )
+    return rules
 
 
 def echo_issues(
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -21,7 +21,7 @@ from typing import (
 import git
 
 from tartufo import config, types, util
-
+from tartufo.types import Rule
 
 BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
 HEX_CHARS = "1234567890abcdefABCDEF"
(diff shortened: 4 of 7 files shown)


tartufo/config.py
```diff
@@ -22,11 +23,11 @@
 import truffleHogRegexes.regexChecks
 
 from tartufo import types, util
-
+from tartufo.types import Rule
 
 OptionTypes = Union[str, int, bool, None, TextIO, Tuple[TextIO, ...]]
 
-DEFAULT_REGEXES = truffleHogRegexes.regexChecks.regexes
+DEFAULT_REGEXES = util.convert_regexes_to_rules(truffleHogRegexes.regexChecks.regexes)
```
tarkatronic, line 30: This has got me hesitant; side-effects at import time should be avoided at all costs. The likelihood of this causing problems is low, but still as a general rule they should be avoided. Is there some other place this initialization code could be placed?

Alternatively, perhaps this shows that #5 needs to be implemented before this, so that we can simply update the data structures we're using, rather than having to manipulate them here.

tartufo/config.py
```diff
@@ -195,22 +196,34 @@ def configure_regexes(
...
     except json.JSONDecodeError as exc:
         raise ValueError(
             "Error loading rules from file: {}".format(rules_file.name)
         ) from exc
-    for rule in new_rules:
-        regexes[rule] = re.compile(new_rules[rule])
-    return regexes
+    for rule_name in new_rules:
+        rule_definition = new_rules[rule_name]
+        if isinstance(rule_definition, collections.Mapping):
```
tarkatronic, line 214: In Python it's generally considered more idiomatic, or "Pythonic", to follow an [EAFP](https://docs.python.org/3.5/glossary.html#term-eafp) (Easier to Ask for Forgiveness than Permission) style as opposed to this [LBYL](https://docs.python.org/3.5/glossary.html#term-lbyl) (Look Before You Leap) style. One major reason being that by checking for one specific type, even a base type like this, you defeat Python's duck typing.

A more idiomatic way to do this check would be something like:

```python
try:
    path_pattern = rule_definition.get("path_pattern", None)
    # ... create your rule
except AttributeError:  # The "get" will raise an AttributeError if this is not a dictionary
    # ... create your rule
```

tarkatronic: As a side note, this also eliminates the need to import `collections`, since the module isn't used elsewhere in here.

tartufo/config.py
```diff
@@ -195,22 +196,34 @@ def configure_regexes(
...
         new_rules = json.load(rules_file)
     except json.JSONDecodeError as exc:
         raise ValueError(
             "Error loading rules from file: {}".format(rules_file.name)
         ) from exc
-    for rule in new_rules:
-        regexes[rule] = re.compile(new_rules[rule])
-    return regexes
+    for rule_name in new_rules:
+        rule_definition = new_rules[rule_name]
```
tarkatronic, line 213:
```suggestion
    for rule_name, rule_definition in new_rules.items():
```
This isn't an absolutely necessary change, but a nice short-hand and common pattern.

tartufo/types.py
```diff
@@ -39,6 +39,13 @@ class Chunk:
     metadata: Dict[str, Any] = field(default_factory=dict)
 
 
+@dataclass
+class Rule:
```
tarkatronic, line 43: I'm wondering if it makes sense to use a more explicit name here. In the scanners, we call these `regex_rules`, which is more of a legacy hold-over than anything else. I'm just concerned that `Rule` is very generic and could be easily misconstrued. What about something like... `RegexRule` (keeping with the previous naming), `MatchRule`... those are the only ones coming to mind right now.

tartufo/util.py
```diff
@@ -36,6 +37,15 @@ def del_rw(_func: Callable, name: str, _exc: Exception) -> None:
     os.remove(name)
 
 
+def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
+    rules: Dict[str, Rule] = {}
+    for regex_name in regexes:
+        rules[regex_name] = Rule(
+            name=regex_name, pattern=regexes[regex_name], path_pattern=None
+        )
+    return rules
```
tarkatronic, line 46: Also not a necessary change here, but another common Python pattern that could be used in this instance: [dictionary comprehensions](https://www.python.org/dev/peps/pep-0274/)!

```python
return {name: Rule(name=name, pattern=pattern, path_pattern=None) for name, pattern in regexes.items()}
```

6 lines down to 1 (as long as `black` keeps it that way...), and temporary variables eliminated!

tarkatronic (dismissed):
This is a great start, thanks @dclayton-godaddy! Just a few suggestions here and we can get this out the door. Although I do have one question in here about whether we should implement #5 first. It might be a bit of a chicken-and-egg scenario.

tartufo/config.py
dclayton-godaddy: In general, I agree. I try to avoid initializing code at the module level (other than constants), however this was already being done by the trufflehog regex module itself and rather than making too extreme changes and the little risk, this makes sense until a refactor is done. https://github.com/dxa4481/truffleHogRegexes/blob/master/truffleHogRegexes/regexChecks.py#L5

tartufo/config.py
dclayton-godaddy: In most languages, business logic by Exception is frowned on as it's a more costly operation. I try to reserve Exception handling to unexpected or edge-case scenarios.

tartufo/config.py
dclayton-godaddy: This is another one of those "somehow this code is being reviewed at a higher level of diligence than the original code"... https://github.com/godaddy/tartufo/blob/master/tartufo/config.py#L212

tartufo/types.py
dclayton-godaddy: I didn't like the name `Rule`, however I also don't like the name `Issue`. So I kept with the style of the package. Given the original code referenced it as `rule` or `new_rules` without being "misconstrued", this seemed like it would get little resistance.

```python
for rule in new_rules:
        regexes[rule] = re.compile(new_rules[rule])
```

tartufo/util.py
dclayton-godaddy: Given the urgency of this PR, do you think we have time for changes like this? Seems more like a personal preference.

jgowdy (approved):

tartufo/util.py
mxhenry-godaddy: I don't think this needs to be a blocker on the PR. It is something worth discussing and settling on an agreed pattern for use in the project moving forward.

tarkatronic: Like Mark and I said, this is not a necessary change or a blocker. It's just a suggestion to show a pattern that is commonly used in Python for instances like this. I'm offering up another way of doing this that would be familiar and readily come to mind for a long-time Python developer, but may not be immediately obvious or known to somebody less familiar with the language. :) 

As to timing, I wouldn't think that copy/pasting a suggestion would really have any impact on anything. But again, this is more an offer of another idiomatic approach than a necessary change or hard blocker.

tartufo/config.py
mxhenry-godaddy: I am in favor of keeping the project as Pythonic as possible. This will also contributors with Python backgrounds to more easily grok and contribute to the code.

tartufo/config.py
mxhenry-godaddy: As we continue to onboard more contributors we should endeavor to continuously uplift the code quality, i.e. constantly raise the bar. The original code was done under a specific set of conditions. Since then we have been making a concerted effort to treat this less in that manner and more as the open source project that it is.

tartufo/config.py
tarkatronic: A key here is, "In most languages". What is considered idiomatic and performant in other languages is not necessarily going to hold true for Python. The reverse is true as well. In Python, the EAFP style of coding is common and encouraged. This is in part because exception handling is not as much of a performance issue. Instead it's a mechanism you're meant to use frequently.

tartufo/config.py
mxhenry-godaddy: Based on meeting conversations I'm thinking that #5 needs to be addressed as part of this.

tarkatronic: I think it's best to keep the work for the 2 separate, to keep the PRs smaller and move focused. But I do agree they should be done alongside each other. It's really just the order in which they should be tackled that's in question. Those two issues are the only things currently tagged for v2.1 though, so I'm not sure the order matters all that much. 😄

tartufo/util.py
dclayton-godaddy: 1 line is not always good. What happens when that line gets too long? Also, the current approach is easier to read. And finally, is it really worth this time?

tartufo/config.py
dclayton-godaddy: I agree, however this PR was not intended to uplift code. This follows the original pattern of the code. I don't recommend you include nor mandate contributors to "uplift" existing code. This type of feedback only blocks added value to this tool.

tartufo/util.py
mxhenry-godaddy: I don't believe @tarkatronic 's suggestion was focused on one line but rather on Pythonic approach, per PEP, to the dictionary.

tartufo/config.py
dclayton-godaddy: #5 is totally unrelated to this change and I agree should not be part of this. Keep things small.

tartufo/config.py
dclayton-godaddy: In that case, you might want to submit an issue with the following repos:

- https://github.com/psf/requests/search?q=isinstance
- https://github.com/urllib3/urllib3/search?q=isinstance

Only the first two packages I looked at that are using `isinstance` before operating on an object.

In areas where they use AttributeError, they are using it for backward compatibility (if they change the methods on an object) on an object already verified by `isinstance` (as in https://github.com/psf/requests/blob/75bdc998e2d430a35d869b2abf1779bd0d34890e/requests/cookies.py#L539).

tartufo/config.py
mxhenry-godaddy: I'm not asking for a rewrite of untouched code. Since lines 211-213 are already being modified, why not change them to the new convention? Does it change the readability that much in comparison to the old pattern?

sfeltner-godaddy: I would suggest that any time a contribution touches a section of code, and that code can be improved on, that improvement should take place naturally.  The goal should be for everyone to incrementally improve the code through contributions where possible.  That being said, the expectation is not to rewrite the code every time a PR is submitted, but to simply be a good contribution citizen and help improve the overall quality of the code.

mxhenry-godaddy (approved):

sjacoby-godaddy (approved):
lgtm

Issue #73, opened 2020-09-11
The new implementation is a bit slower than the old


tarkatronic:
## 🐛 Bug Report

When self-scanning `tartufo`, I found that it was taking a bit longer (using `time tartufo` to measure) to run a full scan. In general, it changed from about 13 seconds to about 20 seconds. I did some preliminary testing and wasn't able to find any obvious reasons why. It'd be good to go through at some point with a fine-toothed profiling comb and see what can be optimized in here.

On a fun note, I found that the `@lru_cache` decorators I used did shave about a second off the run time! Yay!

It's worth mentioning that I built the new implementation with the intention of optimizing for memory rather than speed, using things such as generators etc. I don't know if some of this ended up being detrimental to the speed. But may be worth investigating.

## To Reproduce

Run the following commands and you'll be able to see the discrepancy.

```
git checkout v1.x
time tartufo
git checkout master
time tartufo
```

## Expected Behavior

The new version should be a similar speed, if not faster. Certainly not nearly twice as slow.

bsrz:
Seeing the same as described.

Using 1.1.2
```zsh
$ time tartufo --exclude-paths .tartufo/exclude_paths.txt .
Results have been saved in /var/folders/w5/dkpm7qm95s3gbtffn66fwxb00000gn/T/tmp7sodkfkj
tartufo --exclude-paths .tartufo/exclude_paths.txt .  16.10s user 11.53s system 97% cpu 28.389 total
```

Using 2.0.0
```zsh
$ time tartufo scan-local-repo .
tartufo scan-local-repo .  266.86s user 206.12s system 99% cpu 7:55.45 total
```

tarkatronic:
@bsarrazin I hate to ask you to spend more time running these, but do you know if those timings are consistent, with minimal potential outside influences? (Other heavy processes running on the system)

rscottbailey:
I tried this against the tartufo repo itself on my (physical) Debian system, just for grins. Basically the same behavior, and the 2.0.1 times look pretty stable:

1.1.2:
```bash
$ time tartufo

[...spewage clipped...]
real    0m8.190s
user    0m4.614s
sys     0m1.756s
```

2.0.1:
```bash
$ time tartufo scan-local-repo .

real    0m14.562s
user    0m7.017s
sys     0m4.012s
$ time tartufo scan-local-repo .

real    0m15.262s
user    0m7.182s
sys     0m3.919s
$ time tartufo scan-local-repo .

real    0m15.074s
[...]

PR #119, opened 2020-10-20
Fix local scanning

### LINKED ISSUE #118: Local scans fetch remote origin
## 🐛 Bug Report

When using the "scan-local-repo" command, tartufo still tries to fetch remote origin:

`branches = self._repo.remotes.origin.fetch()`

## To Reproduce

`tartufo scan-local-repo .`

Observe that tartufo helps itself to my remotes and attempts to fetch origin

## Expected Behavior

Local repo scans only utilize the local working copy and git database locally.

jgowdy:
This pull request fixes the issue where tartufo always used remote repo scanning even when told to use a local repo.

### DIFF
--- a/tartufo/commands/scan_local_repo.py
+++ b/tartufo/commands/scan_local_repo.py
@@ -19,6 +19,13 @@ from tartufo.scanner import GitRepoScanner, Issue
     "repo-path",
     type=click.Path(exists=True, file_okay=False, resolve_path=True, allow_dash=False),
 )
+@click.option(
+    "--fetch/--no-fetch",
+    is_flag=True,
+    default=True,
+    show_default=True,
+    help="Controls whether or not the remote repo is fetched prior to local scanning",
+)
 @click.pass_obj
 @click.pass_context
 def main(
@@ -28,10 +35,11 @@ def main(
     since_commit: Optional[str],
     max_depth: int,
     branch: Optional[str],
+    fetch: bool,
 ) -> Tuple[str, List[Issue]]:
     """Scan a repository already cloned to your local system."""
     git_options = types.GitOptions(
-        since_commit=since_commit, max_depth=max_depth, branch=branch
+        since_commit=since_commit, max_depth=max_depth, branch=branch, fetch=fetch
     )
     issues: List[Issue] = []
     try:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -464,17 +464,17 @@ class GitRepoScanner(GitScanner):
         try:
             if self.git_options.branch:
                 # Single branch only
+                if self.git_options.fetch:
+                    self._repo.remotes.origin.fetch(self.git_options.branch)
                 unfiltered_branches = list(self._repo.branches)
                 branches = [
                     x for x in unfiltered_branches if x == self.git_options.branch
                 ]
-                if self.git_options.fetch:
-                    self._repo.remotes.origin.fetch(self.git_options.branch)
             else:
                 # Everything
-                branches = list(self._repo.branches)
                 if self.git_options.fetch:
                     self._repo.remotes.origin.fetch()
+                branches = list(self._repo.branches)
         except git.GitCommandError as exc:
             raise types.GitRemoteException(exc.stderr.strip()) from exc
 
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -475,7 +475,7 @@ class GitRepoScanner(GitScanner):
                 raise types.GitRemoteException(exc.stderr.strip()) from exc
         else:
             # Everything
-            branches = self._repo.branches
+            branches = list(self._repo.branches)
             try:
                 if self.git_options.fetch:
                     print("Fetching remote origin (all branches)")
@@ -483,6 +483,7 @@ class GitRepoScanner(GitScanner):
             except git.GitCommandError as exc:
                 raise types.GitRemoteException(exc.stderr.strip()) from exc
 
+        print("Scanning " + str(len(branches)) + " branches")
         for branch in branches:
             diff_index: git.DiffIndex = None
             diff_hash: bytes
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -461,29 +461,23 @@ class GitRepoScanner(GitScanner):
         """
         already_searched: Set[bytes] = set()
 
-        if self.git_options.branch:
-            # Single branch only
-            unfiltered_branches = list(self._repo.branches)
-            branches = [
-                x for x in unfiltered_branches if x.name == self.git_options.branch
-            ]
-            try:
+        try:
+            if self.git_options.branch:
+                # Single branch only
+                unfiltered_branches = list(self._repo.branches)
+                branches = [
+                    x for x in unfiltered_branches if x.name == self.git_options.branch
+                ]
                 if self.git_options.fetch:
-                    print("Fetching remote origin/" + self.git_options.branch)
                     self._repo.remotes.origin.fetch(self.git_options.branch)
-            except git.GitCommandError as exc:
-                raise types.GitRemoteException(exc.stderr.strip()) from exc
-        else:
-            # Everything
-            branches = list(self._repo.branches)
-            try:
+            else:
+                # Everything
+                branches = list(self._repo.branches)
                 if self.git_options.fetch:
-                    print("Fetching remote origin (all branches)")
                     self._repo.remotes.origin.fetch()
-            except git.GitCommandError as exc:
-                raise types.GitRemoteException(exc.stderr.strip()) from exc
+        except git.GitCommandError as exc:
+            raise types.GitRemoteException(exc.stderr.strip()) from exc
 
-        print("Scanning " + str(len(branches)) + " branches")
         for branch in branches:
             diff_index: git.DiffIndex = None
             diff_hash: bytes
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -461,22 +461,35 @@ class GitRepoScanner(GitScanner):
         """
         already_searched: Set[bytes] = set()
 
-        try:
-            if self.git_options.branch:
-                branches = self._repo.remotes.origin.fetch(self.git_options.branch)
-            else:
-                branches = self._repo.remotes.origin.fetch()
-        except git.GitCommandError as exc:
-            raise types.GitRemoteException(exc.stderr.strip()) from exc
+        if self.git_options.branch:
+            # Single branch only
+            unfiltered_branches = list(self._repo.branches)
+            branches = [
+                x for x in unfiltered_branches if x.name == self.git_options.branch
+            ]
+            try:
+                if self.git_options.fetch:
+                    print("Fetching remote origin/" + self.git_options.branch)
+                    self._repo.remotes.origin.fetch(self.git_options.branch)
+            except git.GitCommandError as exc:
+                raise types.GitRemoteException(exc.stderr.strip()) from exc
+        else:
+            # Everything
+            branches = self._repo.branches
+            try:
+                if self.git_options.fetch:
+                    print("Fetching remote origin (all branches)")
+                    self._repo.remotes.origin.fetch()
+            except git.GitCommandError as exc:
+                raise types.GitRemoteException(exc.stderr.strip()) from exc
 
-        for remote_branch in branches:
+        for branch in branches:
             diff_index: git.DiffIndex = None
             diff_hash: bytes
             curr_commit: git.Commit = None
             prev_commit: git.Commit = None
-
             for curr_commit, prev_commit in self._iter_branch_commits(
-                self._repo, remote_branch
+                self._repo, branch
             ):
                 diff_index = curr_commit.diff(prev_commit, create_patch=True)
                 diff_hash = hashlib.md5(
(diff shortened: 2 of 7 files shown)


jgowdy:
Draft until I fix my tests and code coverage

rfox-godaddy:
This is much better behavior.   Currently, scan-local-repo is reaching out to the git remote, and scanning every branch that exists remotely.   This change does not contact the remote, and (by default) scans all branches that exist locally.

jgowdy:
I'm not sure why the two tests I've added aren't creating the code coverage that I feel they should be, according to codecov.  Any assistance with that would be appreciated.

jcrugzz (approved):
Great work investigating and finding the issue @jgowdy and all others who were involved. Looks good to me overall. I'll let some python pros comment on it with any nits.

gkhurana-godaddy (approved):

tarkatronic:
Thanks for contributing this fix, @jgowdy! This "feature" has been in the back of my mind for some time now.

I haven't taken the time to review the code yet, since I'm on vacation at the moment and only have internet access on my phone. But I wanted to make a few quick notes for any other reviewers.

1. This behavior has existed since "the dawn of time"; i.e., it did and does exist in `truffleHog`.
2. With `1` being the case, it leads me to believe that it is potentially intentional.
3. In some (potentially many) cases, this could be desirable. I personally like the behavior, as it essentially updates your local copy and ensure you are scanning the latest code.
4. With `3` being the case, this should become optional behavior. Perhaps with a `--refresh`/`--no-refresh` flag? Name is up for debate. Fetch or update would be good alternative verbs.
5. This project adheres meticulously to SemVer versioning, meaning that default behavior should not change between bugfix or minor versions.
6. With `6` being the case, this should default to current behavior, until such time as a 3.0 version is released. I imagine that new version will end up changing a number of default options.

Okay maybe not so quick. Hope that all sounds good to everybody! Thank you again for the contributions! It's great to see the engagement picking up here!

jgowdy:
I believe that scanning the remote branches when asked to scan a local repo / checkout violates the "principle of least surprise."  Local means local.  And it doesn't just fetch and then scan locally, it fetches and then scans remotely as far as I can tell.  It makes a local repo nothing more than an indirect pointer to remote origin, which seems rather silly.  I could see an option to have tartufo fetch remote branches before local scanning, for branches that are checked out locally.  If branches aren't checked out locally I wouldn't expect them to be scanned.  But no matter what, the scan should take place locally if the user says scan local repo.  There's no excuse for the scan to be against remote branches when the user says local.

jgowdy:
If semver does require a v3 release then I honestly believe we should consider an expedited 3.0 release cycle.  The behavior of 2.0 isn't rational IMO.

tartufo/scanner.py
```diff
@@ -462,21 +462,32 @@ def chunks(self) -> Generator[types.Chunk, None, None]:
...
+                        x
+                        for x in unfiltered_branches
+                        if x.name == self.git_options.branch
+                    ]
+                else:
+                    branches = self._repo.branches
             else:
-                branches = self._repo.remotes.origin.fetch()
+                if self.git_options.branch:
+                    branches = self._repo.remotes.origin.fetch(self.git_options.branch)
```
jgowdy: Near as I can tell, this is returning a list of Head objects that point to remote, rather than just the list of fetched Heads locally.  Thus I would expect the scan is taking place on the remote branches rather than the local fetched copies.  Otherwise for an already fetched repo the performance difference wouldn't really exist.

jgowdy:
https://github.com/godaddy/tartufo/pull/119#discussion_r509328226

jgowdy (commented):
Looking at the existing code

jgowdy:
Thanks to @djlarsu for helping me debug this and getting my Python code sorted out.

c-jcarlsen-godaddy:
Very nice change and performance gains!

I believe that the previous behavior was intentional, so that the local repo would act as a pointer to the remote repo. This would put the responsibility to fetch history with sufficient depth (tartufo option `max-depth`, on tartufo, and not on the user when cloning the repo. For certain processes that by default clone using shallow clones with depths of 1-10 (Jenkins jobs, GitHub Actions runners) this could potentially be unwanted and lead to tartufo passing in cases where it shouldn't.

I like the idea to allow `scan-local-repo` to accept an option `--fetch` to go fetch the branches, and run it completely local if this option is not provided. 

One could argue that for this we could just use `scan-remote-repo`, but I thinks it's a slightly different usage pattern. In this scenario we have to get and configure the remote url in the `tartufo` command, which seems like extra steps when we already have a nice big `.git` directory in our pwd to point us to the right place to look.

mxhenry-godaddy:
Thoughts after reading through the comments:
To deal with both the performance issue in 2.0.2 while keeping SemVer in mind can we do the following?

1. Add a `no-fetch` flag to `scan-local-repo` which defaults to `false` to keep current behavior
    a. Add a warning for the user to let them know when/if the local repo is stale if they set `no-fetch` to `true`?
2. Add a Feature Issue to change the default behavior of that flag in Milestone 3.0
3. Add notes for how this flag behaves, why it has the default it does, and the plan for it to the documentation

c-jcarlsen-godaddy:
@mxhenry-godaddy To avoid the double negation, could we do `--fetch/--no-fetch` with `--fetch` the default?

tarkatronic:
> @mxhenry-godaddy To avoid the double negation, could we do `--fetch/--no-fetch` with `--fetch` the default?

Yeah if you look at the other options like `--json`/`--no-json`, boolean flags like this are handled pretty easily by `click`. 🙂

As far as SemVer, since this is adding a new feature, it would be a minor version bump (2.1.0). Given that this behavior has potentially always existed, I am not sure this is really fixing the performance bug. We are just adding a new feature which, as a side effect, can speed up execution. So I do think #73 needs to stay open still!

Also a side note: I think it'd be great if we could put together a "best practices" doc for tartufo, separate from this, with recommendations such as specifying `--branch` in CI, to only scan the branch that changed.

mxhenry-godaddy:
Given the testing that @jgowdy and @rfox-godaddy did this seems like it directly addresses the performance degradation in 2.0 (from 1.0) so I do view it as a bugfix for that.

mxhenry-godaddy:
I do agree that if the code has always been doing the remote then the root source of the slowdown is something else

jgowdy:
As far as fetching remote repos and then scanning locally, that makes sense to me.  As far as just using the local repo as a pointer to the remote repo, and then scanning the remote repo, that does not make sense to me.

jgowdy:
As @mxhenry-godaddy pointed out:

https://tartufo.readthedocs.io/en/v2.0.1/usage.html#tartufo-scan-local-repo

```
scan-local-repo
Scan a repository already cloned to your local system.
```

I believe this makes a strong case for the code not honoring the documented behavior.

jgowdy:
My assertion is that this line:

`branches = self._repo.remotes.origin.fetch()`

It not only fetches the remote, but it returns an Iterable list of Head objects, which I believe are *remote* Head references, rather than the local references to the fetched remote data.  Thus despite fetching the remote to local, the scan still takes place against the remote.  Maybe I'm wrong, but I believe this to be the case.

jgowdy:
According to this:

https://gitpython.readthedocs.io/en/stable/reference.html#git.remote.Remote.fetch

```
Returns: | IterableList(FetchInfo, …) list of FetchInfo instances providing detailed information about the fetch results
```

https://gitpython.readthedocs.io/en/stable/reference.html#git.remote.FetchInfo

```
info = remote.fetch()[0]
info.ref           # Symbolic Reference or RemoteReference to the changed
                   # remote head or FETCH_HEAD
```

fetch() returns a FetchInfo, which is an object containing a *RemoteReference* to the fetched head.  Thus when we take this return value as our `branches` reference, this is why I believe we are scanning the remote rather than scanning the now fetch updated local.

tartufo/scanner.py
```diff
@@ -462,21 +462,32 @@ def chunks(self) -> Generator[types.Chunk, None, None]:
         already_searched: Set[bytes] = set()
 
         try:
```
tarkatronic, line 464: I think it makes sense to move this try block to only encapsulate the new `else` clause below, since it's meant to capture errors interacting with remotes (raising `types.GitRemoteException`). Do we know if just checking `self._repo.branches` ends up calling the `git` command? If not, I don't think this particular `try`/`except` is valuable to wrap this code. If so, we probably just need to do eventually raise a slightly different exception for this new behavior.

Hopefully that rambling made some sense. 😄

tests/test_scan_local_repo.py
```diff
@@ -28,6 +28,27 @@ def test_scan_exits_gracefully_when_target_is_not_git_repo(self):
...
+            )
+            result = runner.invoke(cli.main, ["scan-local-repo", repo_path.as_posix()])
+        self.assertLess(result.exit_code, 2)
+
+    def test_scan_branch(self):
+        runner = CliRunner()
+        with runner.isolated_filesystem() as dirname:
+            repo_path = util.clone_git_repo(
+                "https://github.com/godaddy/tartufo.git", target_dir=dirname
+            )
```
tarkatronic: I'm very much not a fan of this, for a couple of reasons. This changes our test suite from being pretty strict unit tests, to relying on external services. It's also going to greatly increase the run time for the test suite, since we're now running two full scans of this project on every test run. This was something that was causing pretty big problems before and actually just recently got fixed (#55).

I think that we should instead use some well thought-out mocks and assertions on those mocks. I would be happy to work with you on setting those up if you would like. 😄

tartufo/commands/scan_local_repo.py
```diff
@@ -31,7 +31,7 @@ def main(
 ) -> Tuple[str, List[Issue]]:
     """Scan a repository already cloned to your local system."""
     git_options = types.GitOptions(
-        since_commit=since_commit, max_depth=max_depth, branch=branch
+        since_commit=since_commit, max_depth=max_depth, branch=branch, local_only=True
```
tarkatronic: We should expose this as an option to the user so that we can maintain the same default behavior as discussed in the other comments. With all the other work you've done it should just be a matter of adding another `@click.option('--fetch/--no-fetch', ...)` similar to what can be seen in `cli.py`. Then pass that bool in to this enum. (Or the negation of that bool rather, since `--no-fetch` would mean `local_only=True`. Unless you feel like all the effort of switching to using a verb-style option instead of descriptive. That's not necessary though.)

tarkatronic (changes requested):
Hey @jgowdy, thanks for putting this together! I've finally had the chance to dig into this code. I really like this feature addition, and I would love to work with you to get this merged soon. If you have any questions about my review, feel free to ping me here or on Slack!

tartufo/scanner.py
```diff
@@ -461,22 +461,36 @@ def chunks(self) -> Generator[types.Chunk, None, None]:
...
+            branches = list(self._repo.branches)
+            try:
+                if self.git_options.fetch:
+                    print("Fetching remote origin (all branches)")
+                    self._repo.remotes.origin.fetch()
+            except git.GitCommandError as exc:
+                raise types.GitRemoteException(exc.stderr.strip()) from exc
+
+        print("Scanning " + str(len(branches)) + " branches")
+        for branch in branches:
```
tarkatronic: This whole section looks to me like it could be condensed back down to the original form and eliminate a whole lot of duplicate code. Unless I'm missing something?

Also, the print calls have got to go!

tarkatronic (changes requested):
In addition to the change requested below, it looks like we're missing any tests for `fetch=False`.

tartufo/scanner.py
```diff
@@ -463,20 +463,28 @@ def chunks(self) -> Generator[types.Chunk, None, None]:
...
+                    x for x in unfiltered_branches if x == self.git_options.branch
+                ]
+                if self.git_options.fetch:
+                    self._repo.remotes.origin.fetch(self.git_options.branch)
             else:
-                branches = self._repo.remotes.origin.fetch()
+                # Everything
+                branches = list(self._repo.branches)
+                if self.git_options.fetch:
+                    self._repo.remotes.origin.fetch()
```
rdrey, line 476: We should probably invert these blocks and first fetch, then list `_repo.branches` in case that list is now longer, right?

rdrey: We should probably also iterate over the remotes instead of hard-coding origin, or have a list of remotes to fetch as a parameter, but that could be a different PR?

jgowdy: I agree, I'll reverse the order of fetch and branch listing.

As far as not hard coding origin, I strongly agree but I'll make that a different PR since that's the existing behavior prior to this PR.

tarkatronic: Yeah, the assumption of `origin` is a bug that has yet to be filed...

rdrey (approved):

tests/test_git_repo_scanner.py
```diff
@@ -150,6 +155,61 @@ def test_all_branches_are_scanned_for_commits(
...
+        self.git_options.branch = "bar"
+        mock_repo.return_value.remotes.origin.fetch.return_value = ["foo", "bar"]
+        mock_repo.return_value.branches = ["foo", "bar"]
+        test_scanner = scanner.GitRepoScanner(
+            self.global_options, self.git_options, "."
+        )
+        mock_iter_commits.return_value = []
+        for _ in test_scanner.chunks:
+            pass
+        mock_iter_commits.assert_has_calls((mock.call(mock_repo.return_value, "bar"),))
```
tarkatronic, line 195: I think the important assertion in this and the test below is going to be `mock_repo.return_value.remotes.origin.fetch.assert_not_called()` as opposed to these assertions on `iter_commits`.

Fix that and I think this is good to go! 👍

tarkatronic (approved):

PR #122, opened 2020-10-26
Switch to pygit2

### LINKED ISSUE #66: Consider switching away from GitPython

`GitPython` is "[in maintenance mode](https://github.com/gitpython-developers/GitPython#development-status)", and has no active development or even active bugfixing. It [does not play well with Windows](https://github.com/gitpython-developers/GitPython/issues/525). And it's actually mostly a wrapper around the `git` command line application. Which seems... strange. 

I would like to see us move away from using `GitPython`, with the hopes of adding first-class Windows support and improving some of the performance of this application.

One option is for us to take over maintenance of `GitPython`. Honestly, that's just not realistic. So let's scratch that one immediately.

A larger discussion around this can be found [in the `dvc` project](https://github.com/iterative/dvc/issues/2215). A lot of good information on the topic can be found in that thread, along with two alternatives:

* [pygit2](https://github.com/libgit2/pygit2)
* [dulwich](https://github.com/dulwich/dulwich)

I am leaning toward `dulwich` at the moment. It's pure Python, and looks a little higher level than `pygit2`. In general I'm okay with low level APIs, but it seems to me that `pygit2` is just a bit too low level/primitive. For example, [this is how you clone a repo over SSH](https://www.pygit2.org/recipes/git-clone-ssh.html). I would rather not have to get that far into the weeds.

This change should be transparent to the user, with the only changes being actual Windows support and hopefully improved performance.

jgowdy:
Considerable performance improvements over both v1.1.2 and v2.0.1 by switching from GitPython to pygit2

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,6 @@ repos:
   - id: pylint
     name: pylint
     entry: pylint
-    args: [--extension-pkg-whitelist=pygit2]
     language: system
     types: [python]
     exclude: ^docs/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,7 @@ repos:
   - id: pylint
     name: pylint
     entry: pylint
+    args: [--extension-pkg-whitelist=pygit2]
     language: system
     types: [python]
     exclude: ^docs/
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -17,7 +17,7 @@ tartufo.scanner
 .. automodule:: tartufo.scanner
    :inherited-members:
    :members:
-   :private-members: _iter_diff_index, _iter_branch_commits
+   :private-members: _iter_diff, _iter_branch_commits
    :show-inheritance:
    :undoc-members:
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -38,15 +38,15 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "20.2.0"
+version = "20.3.0"
 description = "Classes Without Boilerplate"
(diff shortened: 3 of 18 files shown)


jgowdy:
Still need to do the auth hook for remotes

jgowdy:
Remotes (ssh, https, and local+fetch) are working, but the ssh credentials logic needs to allow for an option to specify the location of the files.

Unit tests haven't been fixed up for pygit2

rdrey:
Just wanted to note that this (#122) and #66 are currently blocked by the wheels for https://github.com/libgit2/pygit2/issues/1024 . `pygit2` for `pypy3` also doesn't build in the github actions at the moment. (We'd need to install the build-time dependencies of `pygit2`.) I got the windows tests to pass in my fork.

tarkatronic:
@jgowdy I had a realization this morning. Really more of an "omg DUH" moment. This is a case to listen to the Zen of Python:

> Practicality beats purity.

Why go to all this trouble of handling all of these different protocols for cloning, and trying to fight with another package to get supported added for them cross-platform, when we already have a perfectly working cloning mechanism?

Let's either look at copying over the cloning code from `GitPython` (with attribution, of course), or go with the less-than-ideal option of simply leaving the `GitPython` dependency in there, solely for this one usage.

rdreyer-godaddy:
I like that idea a lot. Clone with GitPython, scan with pygit2. By cloning with GitPython we respect the user's git settings / use the OS' auth setup. By scanning with pygit2 we can ignore all of the user's settings which could otherwise mess with the scan (like the default behaviour in #179).

jgowdy:
> Let's either look at copying over the cloning code from `GitPython` (with attribution, of course), or go with the less-than-ideal option of simply leaving the `GitPython` dependency in there, solely for this one usage.

This makes a great deal of sense!

tarkatronic:
This has been re-implemented in #233.

PR #141, opened 2020-11-09
Fix #140: Ensure a valid output folder name in Windows

### LINKED ISSUE #95: Expand CI to run across multiple platforms

Recently, I've been doing most of my dev work for `tartufo` on a Linux machine. Everything was great and hunky-dory, until I moved over to my Mac and tried running the test suite there. That was when #94 reared its ugly head. We should be able to safe-guard against these types of things happening.

The testing matrix for tartufo should expand to run on as many platforms as we can make it. We should be able to, at minimum, run on both Ubuntu and Mac OS. I believe it is also possible to run GitHub Actions against Windows, but I am uncertain if our underlying `gitpython` dependency supports Windows. It's probably worth trying out though! Ultimately, Windows support would be great!

The only alternative I can see is manually testing on multiple platforms. And that's just not great.

Nothing should be necessary!

### LINKED ISSUE #140: Tartufo showing error on window machine
Hi Team

We have tried below command to generate tartufo output in file ( print output in file )

tartufo -od "tartufo_report" scan-local-repo .

After execute command, we are getting error ( check attached screen shot for reference ).  I think error occurs, because it is trying to create folder “tartufo-scan-results-2020-11-02T16:54:52.950660” on window machine but Colon ( : ) not allowed in folder name on window.

Can you please help us on this point ?  let me know if you have any alternate solution for this issue ?

tartufo window issue
Tartufo GIT repo

tarkatronic:

* [X] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [X] CI related changes
* [X] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking

Fixes #140, #95 

## What's new?

* Fixed an issue where using the `--output-dir` option would error our in Windows due to a `:` in the folder name
* Expanded CI to run across Linux, Windows, and MacOS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   tox:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: [3.6, 3.7, 3.8, 3.9, pypy3]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,9 @@ jobs:
       uses: actions/setup-python@41b7212
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Install Enchant
+      run: brew install enchant
+      if: ${{ matrix.os == 'macos-latest' && matrix.python-version == '3.7' }}
     - name: Install dependencies
       run:  pip install poetry tox codecov tox-gh-actions
     - name: Run tox environments
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,12 +8,14 @@ on:
 
 jobs:
   tox:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: [3.6, 3.7, 3.8, 3.9, pypy3]
+    runs-on: ${{ matrix.os }}
     env:
       PYTHON: ${{ matrix.python-version }}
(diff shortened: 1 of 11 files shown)


sjacoby-godaddy (approved):

rfox-godaddy (approved):

sjacoby-godaddy (approved):

rfox-godaddy (approved):

sjacoby-godaddy (approved):
overall this looks good, i think I'd like to see the windows tests working in a future release instead of being skipped (especially the race condition/perms one)

tarkatronic:
> overall this looks good, i think I'd like to see the windows tests working in a future release instead of being skipped (especially the race condition/perms one)

Agreed, I just didn't want to keep expanding the scope of this change. To explain a bit further:

The tests skipped because of pathing issues were caused because Windows is truncating the username in the the path to, I believe `RUN~1` vs what we expect which is `runadmin`. Those exact values may not be right, but that's the gist of it. Apparently that problem was fixed in Python 3.8+. So the code is technically working; Windows is just being a PITA and causing the tests to fail.

The race condition/permission problem is also specific to the testing conditions. And I agree that it should absolutely be fixed. As far as I can tell, what is happening is that there is some weird thing going on where a process is holding a folder open when we try to remove it. IIRC this looked to me like another test-specific issue. It just seemed like another thing far outside the original scope of this PR.

So yeah I would love to see all of these things actually really fixed. But, Windows is being Windows. 😄

rfox-godaddy (approved):

PR #157, opened 2021-01-06
Add success message and exclusions to output

### LINKED ISSUE #42: Report output on clean or successful scan

### LINKED ISSUE #43: Report out of the list of exclusions

c-jcarlsen-godaddy:

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
closes #42 and closes #43 

## What's new?
- Two new option flags are added: `-q/--quiet` and `-v/--verbose`
- Default output on successful runs using default options:
```
Time: 2021-01-06T14:43:25.090314
```
This output can be suppressed with a new option `-q/--quiet` flag 

- Excluded paths and signatures are reported to the json output (always) and to the console if you provide a `-v/--verbose` flag. Example output given `tartufo -v scan-local-repo .`:
```
Time: 2021-01-06T14:43:25.090314

Excluded paths:
poetry.lock
pyproject.toml

Excluded signatures:
62f22e4500140a6ed959a6143c52b0e81c74e7491081292fce733de4ec574542
ecbbe1edd6373c7e2b88b65d24d7fe84610faafd1bc2cf6ae35b43a77183e80b
```


c-jcarlsen-godaddy:
Adding some more test coverage to util.py to improve codecov report

c-jcarlsen-godaddy:
Test coverage is 100% for the diff now, PR is ready for review

rfox-godaddy (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+v2.3.0 -  January 2021
```
mxhenry-godaddy: Can we change to 

> vX.Y.Z - TBD

please?

mxhenry-godaddy (approved):

tartufo/cli.py
```diff
@@ -146,7 +146,21 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
...
+    "--quiet/--no-quiet",
+    help="Quiet mode. No outputs are reported if the scan is successful and doesn't find any issues",
+    default=False,
+    is_flag=True,
+)
+@click.option(
+    "-v/ ",
+    "--verbose/--no-verbose",
+    help="Verbose mode. When this flag is provided, tartufo will report excluded paths and signatures"
+    " as well as scan issues",
```
tarkatronic: I think we may want this to be a `count`, rather than a flag, like we see here in the `click` documentation: https://click.palletsprojects.com/en/7.x/options/#counting

The idea here is that, once we switch to using a proper logger, we use this to change the logging level. So for example, by default we would probably be at `logging.ERROR`; with `-v` it would bump up to `logging.WARNING`; `-vv` would go to `logging.INFO`; and `-vvv` would go to `logging.DEBUG`.

The other implication here is that I think the help text should be more generic. Something along the lines of:

> Display more verbose output. Specifying this option multiple times will incrementally increase the amount of output.

tarkatronic: We should also make sure that `-q` and `-v` are mutually exclusive. Perhaps throw an error if both are specified.

tartufo/util.py
```diff
@@ -44,29 +45,49 @@ def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
...
     repo_path: str,
     output_dir: Optional[pathlib.Path],
 ) -> None:
     """Print all found issues out to the console, optionally as JSON.
 
     :param issues: The list of issues to be printed out
-    :param as_json: Whether the output should be formatted as JSON
+    :param options: Global options object
+    :param scanner: ScannerBase contaning excluded paths from config tree
+    "param now: Timestamp for the scan in isoformat (str)
```
tarkatronic:
```suggestion
    :param now: Timestamp for the scan in isoformat (str)
```

tartufo/commands/scan_local_repo.py
```diff
@@ -51,4 +52,4 @@ def main(
         util.fail(f"There was an error fetching from the remote repository: {exc}", ctx)
     except types.TartufoException as exc:
         util.fail(str(exc), ctx)
-    return (str(repo_path), issues)
+    return (str(repo_path), issues, scanner)
```
tarkatronic: I'm hesitant to change the return signature here. I see why you've done it, but my concern is that this change has only been applied to one of the commands. What about the other commands? Don't they need similar changes made?

Changing this signature just has potentially wider-reaching consequences than just this. I want to be sure we're fully aware of them before taking that step.

tartufo/util.py
```diff
@@ -44,29 +45,49 @@ def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
     }
 
 
-def echo_issues(
+def echo_result(
     issues: "List[Issue]",
-    as_json: bool,
+    options: "types.GlobalOptions",
+    scanner: "ScannerBase",
```
tarkatronic, line 49: This makes me wonder. If we're passing in the scanner, why are we will passing in the issues? I believe we should be able to access `scanner.issues` for the same list. IIRC I set it up so that `.issues` is a cached property, which gets populated by `.scan()`. Would that make more sense here?

Also, why pass in the `now` param rather than just generating it inside this function? It seems like an almost unnecessary extra parameter.

tartufo/util.py
```diff
@@ -44,29 +45,49 @@ def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
...
     """
-    if as_json:
+    if options.json:
         output = {
             "project_path": repo_path,
             "output_dir": str(output_dir) if output_dir else None,
+            "excluded_paths": [str(path.pattern) for path in scanner.excluded_paths],
+            "excluded_signatures": [
+                str(signature) for signature in options.exclude_signatures
+            ],
```
tarkatronic, line 68: A note for the future: I had taken the initial feature request as meaning, we should print out what was actually *matched* for the exclusions, rather than the configured patterns. Clearly, that is a FAR more complex problem. But I could see that being another feature we want to consider. And perhaps just as valuable as this, if not more so.

tartufo/util.py
```diff
@@ -44,29 +45,49 @@ def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
...
             "found_issues": [issue.as_dict() for issue in issues],
         }
         click.echo(json.dumps(output))
     else:
-        for issue in issues:
-            click.echo(issue)
+        if not issues:
+            if not options.quiet:
+                click.echo(f"Time: {now}")
+                click.echo("All clear. No secrets detected.")
```
tarkatronic:
```suggestion
                click.echo(f"Time: {now}\nAll clear. No secrets detected.")
```
I see no reason not to keep these messages on a single line. 😄
(This is actually a micro-optimization; @jgowdy has found in some experimentation that the `click.echo` function can actually be a bit slow at times. So consolidating calls to it just seems like a good general best practice.)

tartufo/util.py
```diff
@@ -44,29 +45,49 @@ def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
...
+        else:
+            for issue in issues:
+                click.echo(issue)
+        if options.verbose:
+            click.echo("\nExcluded paths:")
+            for path in scanner.excluded_paths:
+                click.echo(path.pattern)
+            click.echo("\nExcluded signatures:")
+            for signature in options.exclude_signatures:
+                click.echo(signature)
```
tarkatronic: Similar to above it might be good to consolidate these echos; maybe something like
```python
click.echo('\n'.join([path.pattern for path in scanner.excluded_paths]))
```

tarkatronic (changes requested):
First review pass. Loving this work, thank you for this! I think there are a few things we need to clean up, but in general this definitely looks like a good direction I'd like to see incorporated soon.

CHANGELOG.md
c-jcarlsen-godaddy: Will update

tartufo/util.py
c-jcarlsen-godaddy: Thanks, updated this

tartufo/util.py
c-jcarlsen-godaddy: Updated

tartufo/util.py
c-jcarlsen-godaddy: Fixed

tartufo/util.py
c-jcarlsen-godaddy: Removed the `now` parameter and instantiating this timestamp in the function instead. My thought here was to have equality between a directory name for `output_dir` and the timestamp in the output. However, both of these happen after the actual scanning of the code, so should be very similar in time. 
I can't see that issues are accessible by `.issues`, but from `._issues` which seems less like it's intended to be exposed to external members. Letting both `issues` and `scanner` be separate arguments for now, but open to arguments 🧐

tartufo/util.py
c-jcarlsen-godaddy: Yes I see this as a valuable feature as well, but I see this as a natural next step in this direction.

tartufo/cli.py
c-jcarlsen-godaddy: Added validation to catch this: 
```
if options.quiet and options.verbose > 0:
    raise click.BadParameter("-v/--verbose and -q/--quiet are mutually exclusive.")
```

c-jcarlsen-godaddy: Updated `--verbose` to be a `count` with `default=0` and `int` in `GlobalOptions`. Updated help text.

tartufo/commands/scan_local_repo.py
jgowdy: I think there's value in not creating the GitRepoScanner object twice though.

c-jcarlsen-godaddy: Right, to @jgowdy 's point here - the scanner contains the compiled list of all the exclusions, gathered from a traversing the different config options to look for exclusions. I see value in not replicating that logic and instead passing the scanner to the output writer

c-jcarlsen-godaddy:
@tarkatronic Updated per review comments

tartufo/util.py
tarkatronic: https://github.com/godaddy/tartufo/blob/6c29000f63a2b10c457d2e7023a66561c833f2cf/tartufo/scanner.py#L119-L130 -- this can be used to access a cached copy of the issues, which is set here at the end of the scan: https://github.com/godaddy/tartufo/blob/6c29000f63a2b10c457d2e7023a66561c833f2cf/tartufo/scanner.py#L264-L265

tartufo/cli.py
```diff
@@ -158,34 +173,38 @@ def main(ctx: click.Context, **kwargs: config.OptionTypes) -> None:
     also be made to work in pre-commit mode, for scanning blobs of text as a
     pre-commit hook.
     """
+
     options = types.GlobalOptions(**kwargs)  # type: ignore
     ctx.obj = options
+    if options.quiet and options.verbose > 0:
+        raise click.BadParameter("-v/--verbose and -q/--quiet are mutually exclusive.")
```
tarkatronic, line 180: Oh hey, nice exception! I didn't know about this one! I wonder if there are other places we should be using this... (in other PRs, of course.)

tartufo/cli.py
```diff
@@ -158,34 +173,38 @@ def main(ctx: click.Context, **kwargs: config.OptionTypes) -> None:
...
 @main.resultcallback()  # type: ignore
 @click.pass_context
 def process_issues(
     ctx: click.Context,
-    result: Tuple[str, List[scanner.Issue]],
+    result: Tuple[str, scanner.ScannerBase],
     **kwargs: config.OptionTypes,
 ):
-    repo_path, issues = result
+    repo_path, scan = result
```
tarkatronic, line 190: I think I'd prefer if this was called `scanner`, just to be more explicit about what it actually is. Not a blocker though!

tartufo/commands/pre_commit.py
```diff
@@ -1,23 +1,25 @@
...
-    issues: List[Issue] = []
+    scanner = None
     try:
         scanner = GitPreCommitScanner(options, str(repo_path))
-        issues = scanner.scan()
+        scanner.scan()
     except types.ScanException as exc:
         util.fail(str(exc), ctx)
-    return (str(repo_path), issues)
+    return (str(repo_path), scanner)
```
tarkatronic, line 25: I really like this new structure for the commands. I think it reads really well. Thanks for making this happen!

tartufo/commands/pre_commit.py
```diff
@@ -1,23 +1,25 @@
...
+from tartufo.scanner import GitPreCommitScanner
 
 
 @click.command("pre-commit")
 @click.pass_obj
 @click.pass_context
-def main(ctx: click.Context, options: types.GlobalOptions) -> Tuple[str, List[Issue]]:
+def main(
+    ctx: click.Context, options: types.GlobalOptions
+) -> Tuple[str, Optional[GitPreCommitScanner]]:
```
tarkatronic, line 15: I wonder... is this actually `Optional`? It seems to me like we're always returning a scanner. And really we should always be. Or was this the result of `mypy` complaining about the `scanner = None`?

tartufo/util.py
```diff
@@ -44,29 +44,42 @@ def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
...
+    if options.json:
         output = {
             "project_path": repo_path,
             "output_dir": str(output_dir) if output_dir else None,
-            "found_issues": [issue.as_dict() for issue in issues],
+            "excluded_paths": [str(path.pattern) for path in scanner.excluded_paths],
+            "excluded_signatures": [
+                str(signature) for signature in options.exclude_signatures
+            ],
+            "found_issues": [issue.as_dict() for issue in scanner.issues],
```
tarkatronic, line 69: We should probably add another key in here with the scan time, so we have parity of reporting. Maybe... `scan_time`? Or `scanned_at`? `scan_finish`? `finish_time`? I'm not sure. Naming things is hard.

tartufo/util.py
```diff
@@ -44,29 +44,42 @@ def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
...
     else:
-        for issue in issues:
-            click.echo(issue)
+        if not scanner.issues:
+            if not options.quiet:
+                now = datetime.now().isoformat("T", "microseconds")
+                click.echo(f"Time: {now}\nAll clear. No secrets detected.")
+        else:
+            for issue in scanner.issues:
+                click.echo(issue)
```
tarkatronic, line 78:
```suggestion
            click.echo("\n".join(scanner.issues))
```
Would this work? Just referencing back to the previous comments about trying to use `click.echo` sparingly for performance concerns. If there are a lot of issues, I could see this being an actually fairly significant slowdown.

tests/test_cli.py
```diff
@@ -192,6 +192,45 @@ def test_command_exits_with_positive_return_code_when_issues_are_found(
...
+        self, mock_scanner: mock.MagicMock
+    ):
+        mock_scanner.return_value.issues = []
+        runner = CliRunner()
+        with runner.isolated_filesystem():
+            result = runner.invoke(cli.main, ["-q", "scan-local-repo", "."])
+        self.assertEqual(result.exit_code, 0)
+        with runner.isolated_filesystem():
+            result = runner.invoke(cli.main, ["-v", "scan-local-repo", "."])
+        self.assertEqual(result.exit_code, 0)
```
tarkatronic, line 241: This should be split out into two separate tests. I know it's repetitive, but that way we know exactly what's failing if somehow one of these breaks.

tests/test_util.py
```diff
@@ -66,27 +67,109 @@ def test_clone_git_repo_raises_explicit_exception_on_clone_fail(
 
 
 class OutputTests(unittest.TestCase):
+    @mock.patch("tartufo.scanner.ScannerBase")
+    @mock.patch("tartufo.types.GlobalOptions")
```
tarkatronic: Not a blocker, but I actually built out a helper so that you don't have to mock options. In this case you would be able to call:

```python
from tests.helpers import generate_options

# ...
options = generate_options(GlobalOptions, json=True, verbose=0)
```

I tend to lean toward as little mocking as possible. But again, not a blocker. 😄

tests/test_util.py
```diff
@@ -66,27 +67,109 @@ def test_clone_git_repo_raises_explicit_exception_on_clone_fail(
...
+            "ooooooooooooo",
+        ]
+        util.echo_result(mock_options, mock_scanner, "", "")
+        mock_click.echo.assert_has_calls(
+            (
+                mock.call("Time: now:now:now\nAll clear. No secrets detected."),
+                mock.call("\nExcluded paths:"),
+                mock.call("package-lock.json\npoetry.lock"),
+                mock.call("\nExcluded signatures:"),
+                mock.call("fffffffffffff\nooooooooooooo"),
```
tarkatronic, line 129: Scanning over this quickly, I didn't realize this was "\n" and just read it as "noooooooooooo" 🤣 

Well played.

tarkatronic (changes requested):
This is looking great! I think there's only one must-do change here, splitting up a test. Other than that I think these were nice-to-haves.

Thanks for the continued effort to get this done!

tartufo/cli.py
c-jcarlsen-godaddy: `scanner` shadows the import name scanner, so I landed on `scan` as name for a scanner instance

tartufo/commands/pre_commit.py
c-jcarlsen-godaddy: You're right, it's the second one, `mypy` complaining about `scanner = None`

tests/test_cli.py
c-jcarlsen-godaddy: Done

tartufo/util.py
c-jcarlsen-godaddy: Added `scan_time`

tests/test_util.py
c-jcarlsen-godaddy: Updated to use the testing helper for options. Nice.

c-jcarlsen-godaddy:
Updated per review comments

tartufo/commands/pre_commit.py
tarkatronic: Oof. We'll have to figure out how to fix that (in another PR), because this return type is inaccurate, which could lead to potential side-effects elsewhere.

tarkatronic (approved):

mxhenry-godaddy (approved):

PR #166, opened 2021-02-05
add no-fetch to code snippets and note about what it does

burgessa23:

* [ ] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
-
Added a note about the new `--no-fetch` flag, and modified related code snippets in `example cleanup.rst`

### DIFF
--- a/docs/source/examplecleanup.rst
+++ b/docs/source/examplecleanup.rst
@@ -44,7 +44,7 @@ More on this later!)
    Scan your repo!
 
    .. note:: The `--fetch/--no-fetch` option for local scans controls whether the local clone is refreshed before scan.
-   This should be set to `no-fetch` for this process. This is an argument to `scan-local-repo`, not the base `tartufo` command.
+      This should be set to `no-fetch` for this process. This is an argument to `scan-local-repo`, not the base `tartufo` command.
 
    .. code-block:: console
 
--- a/docs/source/examplecleanup.rst
+++ b/docs/source/examplecleanup.rst
@@ -42,8 +42,9 @@ More on this later!)
 #. Use ``tartufo`` to scan your repository and find any secrets in its history!
 
    Scan your repo!
+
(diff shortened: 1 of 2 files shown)


sjacoby-godaddy (approved):

docs/source/examplecleanup.rst
```diff
@@ -42,11 +42,13 @@ More on this later!)
 #. Use ``tartufo`` to scan your repository and find any secrets in its history!
 
    Scan your repo!
+   .. note:: The `--fetch/--no-fetch` option for local scans controls whether the local clone is refreshed before scan.
+   This should be set to `no-fetch` for this process.
```
tarkatronic:
```suggestion

   .. note:: The `--fetch/--no-fetch` option for local scans controls whether the local clone is refreshed before scan.
      This should be set to `no-fetch` for this process.
```
Added a blank line before the directive so it will render properly, and added some indentation to stop Sphinx from complaining. 😄 

Sphinx/rST can be super picky. The only way to be really absolutely certain everything renders right is to build the docs, which can be done with either `tox -e docs` or `make -C docs clean html`

tarkatronic: I wonder if we should also explicitly call out that it's an argument to `scan-local-repo` and not an argument to `tartufo` itself? Just in case people add it in the wrong spot and get confused.

tarkatronic (changes requested):
Thanks for adding this @burgessa23! This definitely made for a tricky set of circumstances.

I just had one update, because this docs system is a beast. Other than that, love it!

tarkatronic (approved):
Boom! 💥 Thank you again, @burgessa23!

PR #192, opened 2021-06-11
feat: add exclude-entropy-patterns

### LINKED ISSUE #87: Regex-based Exclusions

To avoid noise in a scan, we'd like to provide exclusions using regular expressions on the contents of files.

An exclusions file that can contain the regular expression exclusion rules.

For example, to ignore a git hash in a url:

`'([ =''/"]+)[a-z0-9]{40}([/''" ])?$'`

dclayton-godaddy:
Add the ability to exclude entropy patterns. This will allow flagged entropy rules using regular expressions for all files or specific files.

* [x] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
- added `exclude-entropy-patterns` option

### DIFF
--- a/tartufo/config.py
+++ b/tartufo/config.py
@@ -243,3 +243,36 @@ def compile_path_rules(patterns: Iterable[str]) -> List[Pattern]:
         for pattern in stripped
         if pattern and not pattern.startswith("#")
     ]
+
+
+def compile_rule(pattern: str) -> Rule:
+    """
+    Compile pattern string to Rule.
+
+    :param pattern: Rule pattern with {path_pattern}::{pattern}
+    :return Rule: Rule object with pattern and path_pattern
+    """
+    parts = pattern.split("::")
+    if len(parts) == 1:
+        parts = [".*", parts[0]]
+    return Rule(
+        name=None,
+        pattern=re.compile("::".join(parts[1:])),
+        path_pattern=re.compile(parts[0]),
+    )
+
+
+def compile_rules(patterns: Iterable[str]) -> List[Rule]:
+    """Take a list of regex string with paths and compile them into a List of Rule.
+
+    Any line starting with `#` will be ignored.
+
+    :param patterns: The list of patterns to be compiled
+    :return: List of Rule objects
+    """
+    stripped = (p.strip() for p in patterns)
+    return [
+        compile_rule(pattern)
+        for pattern in stripped
+        if pattern and not pattern.startswith("#")
+    ]
--- a/tartufo/config.py
+++ b/tartufo/config.py
@@ -252,14 +252,11 @@ def compile_rule(pattern: str) -> Rule:
     :param pattern: Rule pattern with {path_pattern}::{pattern}
     :return Rule: Rule object with pattern and path_pattern
     """
-    parts = pattern.split("::")
-    if len(parts) == 1:
-        parts = [".*", parts[0]]
-    return Rule(
-        name=None,
-        pattern=re.compile("::".join(parts[1:])),
-        path_pattern=re.compile(parts[0]),
-    )
+    try:
+        path, pattern = pattern.split("::", 1)
+    except ValueError:  # Raised when the split separator is not found
+        path = ".*"
+    return Rule(name=None, pattern=re.compile(pattern), path_pattern=re.compile(path))
 
 
 def compile_rules(patterns: Iterable[str]) -> List[Rule]:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -358,19 +358,18 @@ class ScannerBase(abc.ABC):
                 hex_strings = util.get_strings_of_set(word, HEX_CHARS)
 
                 for string in b64_strings:
-                    self.evaluate_entropy_string(
-                        chunk, issues, string, BASE64_CHARS, 4.5
+                    issues += self.evaluate_entropy_string(
+                        chunk, string, BASE64_CHARS, 4.5
                     )
 
                 for string in hex_strings:
-                    self.evaluate_entropy_string(chunk, issues, string, HEX_CHARS, 3)
+                    issues += self.evaluate_entropy_string(chunk, string, HEX_CHARS, 3)
 
         return issues
 
     def evaluate_entropy_string(
         self,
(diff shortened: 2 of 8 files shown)


dclayton-godaddy:
`ci / tox (windows-latest, 3.6) (pull_request)`fails on main build as well. See <https://github.com/godaddy/tartufo/runs/2805575957?check_suite_focus=true>

tarkatronic:
> `ci / tox (windows-latest, 3.6) (pull_request)`fails on main build as well. See https://github.com/godaddy/tartufo/runs/2805575957?check_suite_focus=true

This is unrelated to your changes. I've been doing some hunting and found https://github.com/tox-dev/tox/issues/2077 and https://github.com/tox-dev/tox/issues/1570. Unfortunately neither of those have a solution. But I have a suspicion as to what might be causing it; I added environment caching yesterday. The run that initially primed the cache passed, but subsequent runs have failed. I don't know why this would only trigger on windows/3.6, but I'm going to try digging into that separately.

With that in mind, I don't think we need to consider that a blocker for this PR.

tartufo/config.py
```diff
@@ -243,3 +243,36 @@ def compile_path_rules(patterns: Iterable[str]) -> List[Pattern]:
...
+    :return Rule: Rule object with pattern and path_pattern
+    """
+    parts = pattern.split("::")
+    if len(parts) == 1:
+        parts = [".*", parts[0]]
+    return Rule(
+        name=None,
+        pattern=re.compile("::".join(parts[1:])),
+        path_pattern=re.compile(parts[0]),
+    )
```
tarkatronic: One way to make this a little easier, instead of having to deal with `parts[0]` and `parts[1:]`, you can actually use a combination of multi-assignment and limiting the split. So you could end up with something like:

```python
try:
    path, pattern = pattern.split("::", 1)
except ValueError:  # Raised when the split separator is not found
    path = ".*"
return Rule(
    name=None,
    pattern=re.compile(pattern)
    path_pattern=re.compile(path)
)
```
I wouldn't call this a hard blocker for the PR, but figured I'd share in case you hadn't seen these features before 😄

tartufo/scanner.py
```diff
@@ -311,18 +358,40 @@ def scan_entropy(self, chunk: types.Chunk) -> List[Issue]:
...
+        :param chars: Characters to calculate score
+        :param min_entropy_score: Minimum entropy score to flag
+        """
+        if not self.signature_is_excluded(string, chunk.file_path):
+            entropy_score = self.calculate_entropy(string, chars)
+            if entropy_score > min_entropy_score:
+                if self.entropy_string_is_excluded(string, chunk.file_path):
+                    self.logger.debug("entropy string %s was excluded", string)
+                else:
+                    issues.append(Issue(types.IssueType.Entropy, string, chunk))
```
tarkatronic: I'm a little hesitant on appending to the list inline here, and not returning anything. That's pretty uncommon behavior in Python, and could easily be missed as unexpected behavior. I think it might be best if this were to return the list of issues it found, and then the caller can do
```python
issues += self.evaluate_entropy_string(...)
```
It's just a bit more explicit and obvious what's happening that way.

But, I do love that this logic is split out! Great call on that! Much better for testing and readability!

tartufo/scanner.py
```diff
@@ -247,6 +265,35 @@ def signature_is_excluded(self, blob: str, file_path: str) -> bool:
             in self.global_options.exclude_signatures
         )
 
+    @staticmethod
+    def rule_matches(rule: Rule, string: str, path: str) -> bool:
```
tarkatronic, line 270: I think this method is probably worth throwing an lru cache around, similar to `should_scan`. I could see it being called many times for the same patterns, so that could save a lot of extra computation.

tarkatronic (changes requested):
This is great! I love the new feature; this will be a huge help to our users.

I just have 2 small changes I flagged (and one that's more of a knowledge share than anything), and then we should be able to get this merged in!

tartufo/scanner.py
dclayton-godaddy: Good idea. Updated

tartufo/config.py
dclayton-godaddy: Thanks for the split limit. Didn't know about that. Updated.

tartufo/scanner.py
dclayton-godaddy: Using lru is causing the tests to fail.

tartufo/config.py
dclayton-godaddy: cc @tarkatronic

tartufo/scanner.py
dclayton-godaddy: cc @tarkatronic

tartufo/scanner.py
dclayton-godaddy: @tarkatronic not sure if you have any suggestions

tarkatronic: This came down to 2 oddities.
1. Apparently the order of decorators mattered. With `lru_cache` before `staticmethod`, I got errors stating:
   > TypeError: the first argument must be callable

   I didn't even want to deal with that one. Then with `staticmethod` before `lru_cache` I got errors stating:
   > TypeError: unhashable type: 'Rule'

   So much easier to deal with!
2. With the second error, I just needed to make `Rule` hashable. This is done by way of adding a "magic method" called `__hash__` as a way to uniquely identify individual rules.

tarkatronic (approved):

rbailey-godaddy (approved):
Looks okay to me.

I spent a few minutes thinking about whether it is proper to limit these exclusions only to entropy-based detection, or if instead they should be allowed to apply to regex-based alerts as well. (i.e. basically `--exclude-patterns` instead of `--exclude-entropy-patterns`) I can't convince myself one way or the other is obviously better, so I'd vote for letting the proposed implementation stand.

sushantmimani (approved):
LGTM 🦅

PR #195, opened 2021-06-21
Fixes #194 - Ignore signatures as matches

### LINKED ISSUE #194: Ignore excluded signatures when scanning

Right now, when a scan is performed, if you have `exclude-signatures` in your config, they will show up (rightly so) as entropy matches. While this is just a side effect of tartufo doing its job very well, it doesn't make for a great user experience, and it means that every user needs to explicitly exclude their config file from scans.

While excluding config files does work, it can be potentially problematic. Take Python projects, for example, where you can configure tartufo via the `pyproject.toml` file. Do we really want users to have to exclude that file, and simply assume that no other config in there could possibly contain secrets? What if we eventually support other config files, like `Cargo.toml` for example? Overall the current solution is not so great.

Since tartufo itself has a definitive list of all signatures to be excluded, it should be easy enough to look for any entropy matches in that list, and not report on them.

The other solutions I considered were:

* Auto-ignore the config file, which is bad for the reasons covered above
* Use the new `--exclude-entropy-patterns` feature to ignore BLAKE2 hashes. But this could very easily be far too wide of a net.

Adoption for this should be automatic. Since this is really a feature to fix a bug, there shouldn't be too much necessary aside from the changelog entry.

tarkatronic:

* [X] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [X] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking

Fixes #194 

## What's new?

- Half bugfix, half feature. Now when an excluded signature in your config file is found as an entropy match, tartufo will realize that and no longer report it as an issue.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,6 +29,7 @@ repos:
   rev: v0.901
   hooks:
   - id: mypy
+    args: [--ignore-missing-imports, --follow-imports=silent]
 - repo: https://github.com/psf/black
   rev: 21.5b2
   hooks:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+vX.Y.Z - TBD
+------------
+
+Features:
+
+* [#194](https://github.com/godaddy/tartufo/issues/194) - Half bugfix, half
+  feature. Now when an excluded signature in your config file is found as an
+  entropy match, tartufo will realize that and no longer report it as an issue.
+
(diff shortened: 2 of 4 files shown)


mmicco-godaddy (approved):
Love it.

rbailey-godaddy (approved):
LGTM

mxhenry-godaddy (approved):

PR #196, opened 2021-06-28
Fix caching in Python 3.6/Windows

tarkatronic:

* [ ] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [X] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?

Attempting to fix caching in Windows CI. If this doesn't work, we'll just go ahead and disable caching for Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,10 +48,12 @@ jobs:
       run: brew install enchant
       if: ${{ matrix.os == 'macos-latest' && matrix.python-version == '3.7' }}
     - name: Install dependencies
-      run:  pip install poetry tox tox-gh-actions
+      run: |
+        pip install -U pip
+        pip install poetry tox tox-gh-actions
       # run:  pip install poetry tox codecov tox-gh-actions
     - name: Run tox environments
-      run: tox
+      run: tox -vv
     # - name: Upload coverage to Codecov
     #   uses: codecov/codecov-action@e156083f13aff6830c92fc5faa23505779fbf649
     #   with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
     - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
     - uses: actions/cache@c64c572235d810460d0d6876e9c705ad5002b353
       name: Cache Poetry & Tox deps
+      if: ${{ !(matrix.os == 'windows-latest' && matrix.python-version == '3.6') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -36,9 +36,10 @@ jobs:
         path: |
           {{ matrix.path }}
           .tox
-        key: ${{ runner.os }}-venvs-${{ hashFiles('**/poetry.lock') }}
+        key: ${{ matrix.os }}-venvs-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
         restore-keys: |
-          ${{ runner.os }}-venvs
+          ${{ matrix.os }}-venvs-${{ matrix.python-version }}
+          ${{ matrix.os }}-venvs
```
rbailey-godaddy, line 43: So I barely understand what this is doing, although I am reverse-engineering it in my mind and understand the intent.

With that proviso, should the order of L41 and L42 be reversed? I am assuming that in the case of key conflicts, the later definition would overwrite an earlier one, and presumably we would want any python-version-specific definitions to replace more generic ones...

Basically clueless here.

rbailey-godaddy (commented):
I provided some feedback, but I am not competent to assess the correctness of the change, so I'll leave that to somebody else unless we are tooooootally desperate. ;-)

.github/workflows/ci.yml
tarkatronic: Nope! 😄 
https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action

tarkatronic:
In the end, I went with skipping the caching on Python 3.6/Windows. `tox` was failing with an exit code of `3221225504`, which translates to `0xC0000005`, or `STATUS_ACCESS_VIOLATION`: `The instruction at 0x%08lx referenced memory at 0x%08lx. The memory could not be %s.`

Reference: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55?redirectedfrom=MSDN

TL;DR: Caching is broken in Python 3.6/Windows, so we just don't do it now.

rbailey-godaddy (approved):
I approve, but I learned most of what I know in this area from @jwilhelm-godaddy while reviewing this PR... ;-) But it is nice to see all of the automated tests passing again now.

sushantmimani (approved):
LGTM 🦅

PR #199, opened 2021-06-29
Break out the linting into their own distinct CI steps

tarkatronic:

* [ ] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [X] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?

The goal here is to make the CI a bit more transparent and split it out some more. Right now, the linters all run under Python 3.7, and they all run on every OS. That's all a bit silly.

Now, they should only get run once and in their own steps so that it's obvious when linting is failing, and exactly which linter as well.

At least that's the hope.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,16 +7,13 @@ on:
     branches: [ main ]
 
 jobs:
-  tox:
+  unit-test:
+    name: Unit Tests
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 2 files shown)


tarkatronic:
Once this is approved, we will want to modify the checks required on a PR before merging this. Then other PRs will need to get updated with this new workflow definition.

sjacoby-godaddy (approved):

rbailey-godaddy (approved):

PR #207, opened 2021-08-09
Fixes #96 - Explicitly handle submodules

### LINKED ISSUE #96: Would be good to have an option to exclude commits related to adding/removing git submodules
## Feature Request
Would be good to have an option to exclude commits related to adding/removing git submodules

When adding submodules, git adds a diff which includes the commit hash of the submodule. similar to following:

```
@@ -0,0 +1 @@\n+Subproject commit 821b2b1037dd11d4afb61e1eed870ca4e2fbf608\n
```

This commit is getting reported by tartufo scan.

an example of issue reported by tartufo:
```
{
  "file_path": "packages/a-submodule-package",
  "matched_string": "821b2b1037dd11d4afb61e1eed870ca4e2fbf608",
  "diff": "@@ -0,0 +1 @@\n+Subproject commit 821b2b1037dd11d4afb61e1eed870ca4e2fbf608\n",
  "signature": "eb27ebd567f1d17149e4cc9b2221490e0d5b651c121f06c22a0364444ce8b28c",
  "issue_type": "High Entropy",
  "issue_detail": null,
  "commit_time": "2019-11-07 15:22:22",
  "commit_message": "add submodule",
  "commit_hash": "e10234695b3cdbd79cc9575111c730cd2863bf3f",
  "branch": "origin/master"
}
```

More information on git submodules can be found here - https://git-scm.com/book/en/v2/Git-Tools-Submodules 

The diff matches the following pattern: 
```@@ -0,0 +1 @@\n+Subproject commit 821b2b1037dd11d4afb61e1eed870ca4e2fbf608\n```

I think we tartufo excludes this particular diff pattern it may solve the issue.
Also look at https://git-scm.com/book/en/v2/Git-Tools-Submodules to check if there is a better alternative.

As a workaround I excluded the submodules paths (Of the directory, not the files within the directories).

tarkatronic:

* [X] Tests (if applicable)
* [X] Documentation (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [X] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking

Fixes #96 

## What's new?

Due to an unusual edge case, it was found that submodules could cause `tartufo` to completely bomb and puke its guts up. Specifically what happened was that a submodule was specified, and then all the files for that submodule were actually added to the git tree, resulting in a somewhat broken repository structure which just confused the crap out of `GitPython`.

So this change does two things:

1. By default, explicitly ignore submodules. This resolves the ask in #96, and feels like the right thing to be doing regardless. If this is not desired behavior, users can also restore previous behavior with `--include-submodules`.
2. If a scenario like the one above is encountered when attempting to parse submodules, we now exit gracefully with an error message suggesting what might be wrong.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v3.3.0
+  rev: v4.0.1
   hooks:
   - id: check-added-large-files
   - id: fix-byte-order-marker
@@ -17,7 +17,7 @@ repos:
     name: tartufo
     entry: tartufo
     args: [pre-commit]
-    language: python
+    language: system
     pass_filenames: false
   - id: pylint
     name: pylint
@@ -26,12 +26,12 @@ repos:
     types: [python]
     exclude: ^docs/
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.901
+  rev: v0.910
   hooks:
   - id: mypy
     args: [--ignore-missing-imports, --follow-imports=silent]
 - repo: https://github.com/psf/black
-  rev: 21.5b2
+  rev: 21.7b0
   hooks:
     - id: black
       language_version: python3.7
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+vX.Y.Z - TBD
+------------
+
+Features:
+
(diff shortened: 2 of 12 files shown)


rbailey-godaddy (approved):
LGTM

sushantmimani (approved):
LGTM!

PR #212, opened 2021-08-19
feat(scanner): add support for scanning folders

### LINKED ISSUE #83: Support Scanning Directory

dclayton-godaddy:
Add support for scanning folders.

```sh
tartufo scan-folder .
```

* [x] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
closes #83 

## What's new?
- New `scan-folder` subcommand for scanning a folder.


dclayton-godaddy:
@tarkatronic Can you take a look at this?

tartufo/scanner.py
```diff
@@ -683,3 +683,49 @@ def chunks(self):
         )
         for blob, file_path in self._iter_diff_index(diff_index):
             yield types.Chunk(blob, file_path, {})
+
+
+class FolderScanner(ScannerBase):
+    """Used to scan a folder."""
+
+    folder_path: str
```
tarkatronic: Did you mean for this to be `target`, to match line 696?

tartufo/scanner.py
```diff
@@ -683,3 +683,49 @@ def chunks(self):
...
+        """Yield the individual file changes currently staged for commit.
+
+        :rtype: Generator[Chunk, None, None]
+        """
+
+        for blob, file_path in self._iter_folder():
+            if self.should_scan(file_path):
+                yield types.Chunk(blob, file_path, {})
+
+    def _iter_folder(self) -> Generator[types.Tuple, None, None]:
```
```suggestion
    def _iter_folder(self) -> Generator[Tuple[str, str], None, None]:
```

tartufo/scanner.py
```diff
@@ -683,3 +683,49 @@ def chunks(self):
...
+        folder_path = pathlib.Path(self.target)
+        for file_path in folder_path.rglob("**/*"):
+            relative_path = file_path.relative_to(folder_path)
+            if file_path.is_file() and self.should_scan(str(relative_path)):
+                with file_path.open("rb") as fhd:
+                    data = fhd.read()
```
tarkatronic: This should probably be in a try/except block to fail gracefully (display an error message?) when we can't open a file.

docs/source/features.rst
```diff
@@ -54,6 +54,31 @@ To use ``docker``:
 When used this way, `tartufo` will clone the repository to a temporary
 directory, scan the local clone, and then delete it.
 
+Scanning a Folder
++++++++++++++++++++++++++++
+
+This is scanning a local folder and does not look at git commit history. Ideal for locating
+secrets in the latest source files or files not in source control.
```
tarkatronic: These sentences read a little bit awkwardly. Could we update them to something a little more friendly, maybe along the lines of

> Operating in this mode, tartufo scans the files in a local folder, rather than operating on git commit history. This is ideal for locating secrets in the latest version of source files, or files not in source control.

tarkatronic (changes requested):
Thanks for putting in the work on this @dclayton-godaddy, this is a great start!

I've added just a couple of very small updates for correctness and readability here. In addition, I think there is one must-have feature before this gets merged. Specifically, if a user runs this `scan-folder` command on a git repository, we need to raise some kind of warning and force the user to explicitly opt-in to this scanning mode. This can help prevent some easy mistakes, and suggest that the user _really should_ be scanning commit history when they're looking at a repository.

I've got 2 ideas for how this could work. I think I'm a bit partial to the first, but I could be convinced otherwise.
1. When the user runs `tartufo scan-folder .` on a git repository, a message is displayed, something like. "This folder is a git repository, and should be scanned using the scan-local-repo command. Are you sure you wish to proceed? [Y/n]"
2. By default, a message like the above is displayed, minus the last sentence, and the command exits. The user can pass in a flag along the lines of "--force-scan-repository" to prevent the message, and explicitly opt in.

I think the second option makes it a little too easy to automate your way out of this, for instance, in a GitHub Action. And the extra flag will just become users' default mode of operation, as opposed to forcing the user to actually confirm every time. But again I could easily go either way.

The one other "nice to have" option I think would be a `--recurse/--no-recurse` flag. But I wouldn't worry too much about that one unless you feel like adding it. I'm not going to block the PR for that. 😄 

Thanks again for working on this! I look forward to being able to add this functionality!

docs/source/features.rst
dclayton-godaddy: I agree. I was matching the wording of the other mode.

```
This is the "classic" use case for ``tartufo``: Scanning the history of a git
repository. There are two ways to invoke this functionality, depending if you
are scanning a repository which you already have cloned locally, or one on a
remote system.
```

tartufo/scanner.py
dclayton-godaddy: Sure. I know tartufo reads many files. Can you point me to an example of one that wraps with a try/except?

dclayton-godaddy: nvm, I found one. Using

```
click.FileError(filename=str(config_path / value), hint=str(exc))
```

dclayton-godaddy: This example unfortunately only handles file not found and not permissions errors. Ok to capture `OSError`?

tarkatronic: Yeah I think you would basically want something like
```python
try:
    with file_path.open("rb") as fhd:
        data = fhd.read()
except OSError as exc:
    log.warning("Could not open file %s. (%s)", file_path, exc)
```

tarkatronic: The `OSError` class serves as the base class for all the exception types you would need to capture here, including permissions, file not found, etc. https://docs.python.org/3/library/exceptions.html#os-exceptions

dclayton-godaddy:
@tarkatronic can you review again? All comments should be addressed. Thanks!

tartufo/scanner.py
```diff
@@ -683,3 +685,51 @@ def chunks(self):
...
+
+        :param global_options: The options provided to the top-level tartufo command
+        :param target: The local filesystem path to scan
+        """
+        self.target = target
+        super().__init__(global_options)
+
+    @property
+    def chunks(self) -> Generator[types.Chunk, None, None]:
+        """Yield the individual file changes currently staged for commit.
```
```suggestion
        """Yield the individual files in the target directory.
```

tartufo/scanner.py
```diff
@@ -683,3 +685,51 @@ def chunks(self):
...
+    def _iter_folder(self) -> Generator[Tuple[str, str], None, None]:
+        folder_path = pathlib.Path(self.target)
+        for file_path in folder_path.rglob("**/*"):
+            relative_path = file_path.relative_to(folder_path)
+            if file_path.is_file() and self.should_scan(str(relative_path)):
+                try:
+                    with file_path.open("rb") as fhd:
+                        data = fhd.read()
+                except OSError as exc:
+                    raise click.FileError(filename=str(file_path), hint=str(exc))
```
tarkatronic, line 727: Are we sure that we want to raise an exception here and abort the scan? Or should we (similar to a utility like `df`) just print out an error message and continue on? I could honestly go either way here.

If we are going to go with the route of aborting, we'll want two quick enhancements. First, just a general best practice (possibly not followed everywhere in the codebase at the moment, but should be) would be to use the `raise ... from` syntax:

```suggestion
                    raise click.FileError(filename=str(file_path), hint=str(exc)) from exc
```

The other is that in the command itself, we will want an additional

```python
except click.FileError:
```
Similar to what is done with `scan-remote-repo`, to make sure we exit gracefully instead of bombing out with a stack trace.

tarkatronic (changes requested):
This is looking great! Just a couple final small tweaks and we should be good to go!

tartufo/scanner.py
dclayton-godaddy: If the file might have a secret, I don't want it to be hidden by log entry. The user should omit the file if they don't want it scanned. As far as the stack trace, I just copied that from other code. I think just showing the exception message as the hint and log the full exception might be better. I agree, I don't like the stack trace. I also thought click handles `FileError` gracefully.

dclayton-godaddy: This is what it would look like:
```
Error: Could not open file /Users/dclayton/git/github.com/dclayton-godaddy/tartufo/tests/data/scan_folder/test.txt: [Errno 13] Permission denied: '/Users/dclayton/git/github.com/dclayton-godaddy/tartufo/tests/data/scan_folder/test.txt'
```

tarkatronic: Ah interesting, there must be some magical internal handling for file errors. That works for me!

docs/source/features.rst
```diff
@@ -54,6 +54,32 @@ To use ``docker``:
...
+
+   $ docker run --rm -v "/path/to/my/repo:/git" godaddy/tartufo scan-folder /git
+
+.. note::
+
+   If you are using ``podman`` in place of ``docker``, you will need to add the
+   ``--privileged`` flag to the ``run`` command, in order to avoid a permission
+   denied error.
+
+   This will scan all files and folders in the current working directory including
```
rbailey-godaddy:
```suggestion
   This will scan all files and folders in the specified directory including
```

tartufo/scanner.py
rbailey-godaddy: I agree it's better to abort the scan rather than obfuscate the fact that a selected target might contain secrets that we didn't report. Plus it's simple. FWIW, it seems the "continue on" alternative would require us to dummy up a finding that represents the entire file (if we can't assess it, we assume it's problematic) and that just doesn't seem like it's worth the effort in this context.

tartufo/scanner.py
```diff
@@ -683,3 +685,51 @@ def chunks(self):
...
+
+                try:
+                    blob = data.decode("utf-8")
+                except UnicodeDecodeError:
+                    # binary file, continue
+                    continue
```
rbailey-godaddy, line 733: I'm not meaning to pick on this scanner in particular, but I'm not sure this is the proper approach. If the target file happens to be (I tried this as a test) a SQLite database that might contain tons of sensitive data, consider:

```python
>>> with open("gerbera.db", "rb") as f:
>>>     x = f.read()
>>> y = x.decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa8 in position 35: invalid start byte
>>> y = str(x)
>>> "Adagio" in x
TypeError: a bytes-like object is required, not 'str'
>>> "Adagio" in y
True
```
Abstractly, I'd be happier if we tried to just look at the bytes (which might or might not be useful -- after all, maybe the input file is in EBCDIC), since in many cases we might in fact be able to see interesting content. The output of a match might be really horrific, so maybe this should just be food for future thought.

rbailey-godaddy (approved):
There's an editorial comment/suggestion I'd prefer to see made (the first comment below) and the other stuff either I agree with or I think is a bigger question that shouldn't necessarily be inflicted on this PR.

tartufo/scanner.py
dclayton-godaddy: Ya. Sounds like a nice feature that should be applied overall. For now, I think we just scan non-binary files.

tarkatronic (approved):

PR #213, opened 2021-08-19
Fixes #211 - Fix windows encoding

### LINKED ISSUE #211: Tartufo scanning issue on window
## 🐛 Bug Report
when trying to scan GIT repo on window machine using ```tartufo scan-local-repo .``` command. i am getting below issue
image

i feel, might be one of the file have UTF-8 characters which tartufo not able to read

## To Reproduce

Add UTF8 character in file and scan it ( specially Chinese characters like 开始搜  )

## Expected Behavior

scanning should complete on Window machine

tarkatronic:

* [ ] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking

Fixes #211

## What's new?

There's a strange case when using git bash on Windows where stdout will be opened with a CP-1252 encoding, as opposed to UTF-8, when it's opened as a text stream. This would cause problems with scanned failed contained Unicode characters adjacent to or as part of matched strings, because CP-1252 can't handle outputting those characters. So now we output the issues as UTF-8 encoded bytestreams, which means that the output is handled in pure binary and we no longer have to worry about the output encoding.

This should be fully backward compatible and theoretically allow proper Unicode output across all environments.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -78,9 +78,9 @@ class OutputTests(unittest.TestCase):
         mock_scanner.exclude_signatures = []
         mock_scanner.issues = [1, 2, 3, 4]
         util.echo_result(options, mock_scanner, "", "")
-        # This is the result of calling `bytes()` on integers.
+        # Ensure that the issues are output as a byte stream
         mock_click.echo.assert_called_once_with(
-            b"\x00\n\x00\x00\n\x00\x00\x00\n\x00\x00\x00\x00"
+            bytes(1) + b"\n" + bytes(2) + b"\n" + bytes(3) + b"\n" + bytes(4)
         )
 
     @mock.patch("tartufo.scanner.ScannerBase")
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -78,7 +78,10 @@ class OutputTests(unittest.TestCase):
         mock_scanner.exclude_signatures = []
         mock_scanner.issues = [1, 2, 3, 4]
         util.echo_result(options, mock_scanner, "", "")
-        mock_click.echo.assert_called_once_with("1\n2\n3\n4")
+        # This is the result of calling `bytes()` on integers.
+        mock_click.echo.assert_called_once_with(
+            b"\x00\n\x00\x00\n\x00\x00\x00\n\x00\x00\x00\x00"
+        )
 
     @mock.patch("tartufo.scanner.ScannerBase")
     @mock.patch("tartufo.util.click")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+vX.Y.Z - TBD
+------------
+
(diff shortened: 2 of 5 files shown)


tests/test_util.py
```diff
@@ -78,7 +78,10 @@ def test_echo_result_echos_all_when_not_json(self, mock_click, mock_scanner):
         mock_scanner.exclude_signatures = []
         mock_scanner.issues = [1, 2, 3, 4]
         util.echo_result(options, mock_scanner, "", "")
-        mock_click.echo.assert_called_once_with("1\n2\n3\n4")
+        # This is the result of calling `bytes()` on integers.
+        mock_click.echo.assert_called_once_with(
+            b"\x00\n\x00\x00\n\x00\x00\x00\n\x00\x00\x00\x00"
+        )
```
rbailey-godaddy: This doesn't thrill me because I can't make out any relationship between the input and output that convinces me we know what is supposed to happen or that the input even has an impact on the output.

What if you go back to the original code (original L79-L81) and change L79 to read

```python
mock_scanner.issues = [b"1", b"2", b"3", b"4"]
```
and then prefix the string in L81 with `b`?

rbailey-godaddy (changes requested):
I think we could improve the unit test change; see comment inline.

tests/test_util.py
tarkatronic: I can see the difficulty of interpreting this at a glance. The byte string we are asserting isn't very visually pleasant to parse. As to whether "we know what is supposed to happen", I would say that this test (still) is direct proof that we know. Specifically this is the results of `byte(1) + "\n" + byte(2) + "\n" + byte(3) + "\n" + byte(4)`.

I had initially tried to make this a little more pleasing to read, but this turned out to be the far easier option. That said, I can put some more effort into making this easier to visually parse.

tarkatronic: Would you find this easier to parse if I did something like this in the assert?
```pycon
>>> bytes(1) + b"\n" + bytes(2) + b"\n" + bytes(3) + b"\n" + bytes(4)
b'\x00\n\x00\x00\n\x00\x00\x00\n\x00\x00\x00\x00'
```

rbailey-godaddy: Yes. Or alternatively add a comment explaining to the reader that we expect the UTF-8 encoding of integer N to be a string of N nul bytes. (Something I "know" only by reverse-engineering the observed output.)

sushantmimani (commented):
Agree with @rbailey-godaddy comment on the unit test. Code changes look good.

sushantmimani (approved):
LGTM 🦅

rbailey-godaddy (approved):

PR #219, opened 2021-09-11
Incremental optimizations

tarkatronic:
This includes just a couple of small, incremental tweaks that should offer a little bit of performance gain.

* Using `__slots__` for the `Issue` class should help a lot during initial deep scans of large repositories. This should improve both memory consumption and, to some lesser degree, the speed of instantiating the Issue class.
* Doing some quick initial profiling, it appears that generating the diff between hashes is one of the most run lines in the code, and also one of the more expensive. So now we don't even generate it until after we know the file should be scanned.

Both of these are pretty minor, but should help some, especially in more extreme cases.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -16,7 +16,7 @@ python-versions = "*"
 
 [[package]]
 name = "astroid"
-version = "2.6.6"
+version = "2.7.3"
 description = "An abstract syntax tree for Python with inference support."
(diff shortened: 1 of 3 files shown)


rbailey-godaddy (approved):
LGTM

sushantmimani (approved):
LGTM 🦅

PR #220, opened 2021-09-13
fix(git): raise error if branch not found

dclayton-godaddy:

* [x] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [x] No (breaking changes)

## What's new?
Currently, if someone runs the tartufo CLI using the `--branch` argument and the branch does not exist, the CLI will return no secrets found. This is misleading and should fail the call. This change will fail the call if `--branch` is specified and the branch was not found.

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -24,7 +24,7 @@ import click
 import git
 
 from tartufo import config, types, util
-from tartufo.types import Rule, TartufoException
+from tartufo.types import BranchNotException, Rule, TartufoException
 
 BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
 HEX_CHARS = "1234567890abcdefABCDEF"
@@ -609,6 +609,11 @@ class GitRepoScanner(GitScanner):
                 branches = [
                     x for x in unfiltered_branches if x.name == self.git_options.branch
                 ]
+
+                if len(branches) == 0:
+                    raise BranchNotException(
+                        f"Branch {self.git_options.branch} was not found."
+                    )
             else:
                 # Everything
                 if self.git_options.fetch:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -24,7 +24,7 @@ import click
 import git
 
 from tartufo import config, types, util
-from tartufo.types import BranchNotException, Rule, TartufoException
+from tartufo.types import BranchNotFoundException, Rule, TartufoException
 
 BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
 HEX_CHARS = "1234567890abcdefABCDEF"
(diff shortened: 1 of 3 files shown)


tartufo/scanner.py
```diff
@@ -609,6 +609,11 @@ def chunks(self) -> Generator[types.Chunk, None, None]:
                 branches = [
                     x for x in unfiltered_branches if x.name == self.git_options.branch
                 ]
+
+                if len(branches) == 0:
+                    raise BranchNotException(
+                        f"Branch {self.git_options.branch} was not found."
+                    )
```
rbailey-godaddy: Well, `BranchNotException` irks me because it feels like we're missing a word... the branch isn't what? Found? Valid?

Worse, I went and looked to see what tartufo was using for a convention, and they're pretty generic:

```
class TartufoException(Exception):
class ConfigException(TartufoException):
class ScanException(TartufoException):
class GitException(TartufoException):
class GitLocalException(GitException):
class GitRemoteException(GitException):
```

This makes me think we probably should be raising `GitLocalException` for a local repo or `GitRemoteException` for a remote repo, but I get that we probably don't want to have to screw around trying to pick the right one.

My suggestion here therefore would be to use `GitBranchException` (which should be a specialization of `GitException`).

The other option that came to mind would be some sort of `UserFootInMouthException` for when we are whining about GIGO directives, but there doesn't seem any call for it at this point so I'd stick with something simple.

tartufo/types.py
```diff
@@ -103,6 +103,10 @@ class ScanException(TartufoException):
     """Raised if there is a problem encountered during a scan"""
 
 
+class BranchNotException(TartufoException):
+    """Raised if a branch was not found"""
```
rbailey-godaddy: Per above, I'd make this `GitBranchException(GitException)` and you might need to move it down if this is ahead of the `GitException` declaration.

rbailey-godaddy (changes requested):
I don't like the proposed name of the exception, but if you can sell @jwilhelm-godaddy on it I'll change my vote. ;-)

tartufo/scanner.py
dclayton-godaddy: I did not want to mark it as a GitException as it wasn't an issue interacting with Git (`Raised if there is a problem interacting with git`). This is a logic issue unrelated to interacting with git. For example, if someone typed in `mybrach` and missed the `n`, this is a user issue where it would say `mybrach not found`.

tarkatronic: Regardless of all else, I agree that `BranchNotException` doesn't particularly make sense. As @rbailey-godaddy said, it sounds like it's missing a word. I would recommend renaming it to `BranchNotFoundException` would be very appropriate.

dclayton-godaddy: lol. You're right. Not sure what I was thinking. I changed to BranchNotFoundException.

tarkatronic (approved):

rbailey-godaddy (approved):

sushantmimani (approved):
LGTM 🦅

PR #227, opened 2021-10-13
Feature: report issues as discovered instead of buffering

rbailey-godaddy:
This is a reimplementation of PR #108 which streams output (including JSON).

* [ ] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?

Issues are now written to the console as soon as they are detected, while the scan is in progress, instead of buffering all of them and reporting only when the scan has completed. This is generally backward compatible:
* A user who starts a scan and walks away, returning only after the scan has completed, will not detect any changes in behavior.
* A caller who references only the `issues` property (and not the `scan()` method) will see no changes in behavior.
* A caller who references only the `scan()` method (and not the `issues` property) will see no changes in behavior, although the method is now a generator that returns issues in succession rather than a function which returns a list. A caller who does `len(scanner.scan())` (or anything else that assumes the result is a list) should use `scanner.issues` (which is a list) instead.
* A caller using both `scan()` and `issues`, where _all_ `scan()` references occur before _all_ `issues` references, will see no changes in behavior.

These cases are not backward compatible: 
* A caller who invokes `scan()` multiple times, expecting to see the scan re-run each time, will observe different behavior. The first series of invocations identifies (and caches) issues, but subsequent invocations re-use the cached issues instead of actually rescanning the target. It looks the same (if you allow for generator vs. list) but you are not actually doing repeated scans.
* A caller who references `issues` and then calls `scan()` will see that the `scan()` call now will iterate over the cached issue list instead of rescanning. Once the caller has referenced the `issues` property, any subsequent `scan()` calls should be replaced with references to `issues`.
* In particular, consuming some (partial) number of issues from `scan()` and then referencing `issues` will return the full list of issues, but cause `scan()` to generate from the cached issues list, meaning that the initial issues will appear to be duplicated. The caller should either use `issues` exclusively, or ensure all references to `issues` occur after all calls to `scan()`.

This change was accomplished using the following strategy:
* `ScannerBase` methods which previously appended to the `_issues` member have been converted to generators which yield individual issues; the lowest-level methods still append to `_issues` (so as to cache a record of all issues) in addition to yielding the issue. In particular, `scan()` is now a generator.
* A new `_completed` member (and matching `completed` property) track whether a scan is still in progress (i.e. whether or not the `scan()` generator is exhausted); this is initially `False` and set to `True` by `scan()` when it finally exits.
* The `issues` property now tests to see if the scan is still in progress, and forcibly drains the generator if needed prior to returning the issues list (which otherwise could be incomplete).
* `echo_output()` now loops on `scan()` instead of `issues` in order to obtain issues individually as they are detected, allowing output to be produced as soon as it is encountered; some code reordering was performed to ensure no references to `issues` occur before all references to `scan()` have been made.
* The output blocks of `echo_output()` were refactored to output individual issues instead of consolidating output to a single large result at the end -- in the most complicated of these, JSON output is handled by generating it in pieces: an initial part, a part for each issue, and a terminal part -- the resulting output is unchanged, but it now appears incrementally;
* Unit testing generally required only minor fixups. The JSON-related unit tests were modified to perform end-to-end validation of the output stream to ensure we're generating valid JSON that correctly reflects the original input data.

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -131,7 +131,7 @@ class ScannerBase(abc.ABC):
     all the individual pieces of content to be scanned.
     """
 
-    _issues: Optional[List[Issue]] = None
+    _issue_count: int = -1
     _included_paths: Optional[List[Pattern]] = None
     _excluded_paths: Optional[List[Pattern]] = None
     _excluded_entropy: Optional[List[Rule]] = None
@@ -144,18 +144,16 @@ class ScannerBase(abc.ABC):
         self.logger = logging.getLogger(__name__)
 
     @property
-    def issues(self) -> List[Issue]:
+    def issues(self) -> Generator[Issue, None, None]:
         """Get a list of issues found during the scan.
 
         If a scan has not yet been run, run it.
 
         :return: Any issues found during the scan.
-        :rtype: List[Issue]
         """
-        if self._issues is None:
+        if self._issue_count < 0:
             self.logger.debug("Issues called before scan. Calling scan now.")
-            self._issues = self.scan()
-        return self._issues
+            yield from self.scan()
 
     @property
     def included_paths(self) -> List[Pattern]:
@@ -345,7 +343,7 @@ class ScannerBase(abc.ABC):
                 entropy += -prob_x * math.log2(prob_x)
         return entropy
 
-    def scan(self) -> List[Issue]:
+    def scan(self) -> Generator[Issue, None, None]:
         """Run the requested scans against the target data.
 
         This will iterate through all chunks of data as provided by the scanner
@@ -355,7 +353,7 @@ class ScannerBase(abc.ABC):
         :raises types.TartufoConfigException: If there were problems with the
           scanner's configuration
         """
-        issues: List[Issue] = []
+
         if not any((self.global_options.entropy, self.global_options.regex)):
             self.logger.error("No analysis requested.")
             raise types.ConfigException("No analysis requested.")
@@ -364,46 +362,43 @@ class ScannerBase(abc.ABC):
             raise types.ConfigException("Regex checks requested, but no regexes found.")
 
         self.logger.info("Starting scan...")
+        self._issue_count = 0
         for chunk in self.chunks:
             # Run regex scans first to trigger a potential fast fail for bad config
             if self.global_options.regex and self.rules_regexes:
-                issues += self.scan_regex(chunk)
+                yield from self.scan_regex(chunk)
             if self.global_options.entropy:
-                issues += self.scan_entropy(
+                yield from self.scan_entropy(
                     chunk,
                     self.global_options.b64_entropy_score,
                     self.global_options.hex_entropy_score,
                 )
-        self._issues = issues
-        self.logger.info("Found %d issues.", len(self._issues))
-        return self._issues
+        self.logger.info("Found %d issues.", self._issue_count))
 
     def scan_entropy(
         self, chunk: types.Chunk, b64_entropy_score: float, hex_entropy_score: float
-    ) -> List[Issue]:
+    ) -> Generator[Issue, None, None]:
         """Scan a chunk of data for apparent high entropy.
 
         :param chunk: The chunk of data to be scanned
         :param b64_entropy_score: Base64 entropy score
         :param hex_entropy_score: Hexadecimal entropy score
         """
-        issues: List[Issue] = []
+
         for line in chunk.contents.split("\n"):
             for word in line.split():
                 b64_strings = util.get_strings_of_set(word, BASE64_CHARS)
                 hex_strings = util.get_strings_of_set(word, HEX_CHARS)
 
                 for string in b64_strings:
-                    issues += self.evaluate_entropy_string(
+                    yield from self.evaluate_entropy_string(
                         chunk, line, string, BASE64_CHARS, b64_entropy_score
                     )
 
                 for string in hex_strings:
-                    issues += self.evaluate_entropy_string(
+                    yield from self.evaluate_entropy_string(
                         chunk, line, string, HEX_CHARS, hex_entropy_score
                     )
-
-        return issues
 
     def evaluate_entropy_string(
         self,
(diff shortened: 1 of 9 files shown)


sushantmimani (approved):
LGTM 🦅

tartufo/scanner.py
```diff
@@ -143,18 +144,30 @@ def __init__(self, options: types.GlobalOptions) -> None:
...
         """
-        if self._issues is None:
-            self.logger.debug("Issues called before scan. Calling scan now.")
-            self._issues = self.scan()
+
+        if not self.completed:
+            self.logger.debug(
+                "Issues called before scan completed. Finishing scan now."
+            )
+            list(self.scan())
```
tarkatronic, line 177: I'm not quite sure about this method of handling this. I'm almost starting to think it's gotta be more of a state machine. Imagine, if you will:

* I am a 3rd party developer using `tartufo` as a library
* In my program, I spin up a new thread to run `scanner.scan()`.
* In my main thread, I have code along the lines of:
    ```python
    while not scanner.issues:
        print("Waiting for scan to complete...")
        time.sleep(5)
    ```
* It seems that this is now going to cause 2 concurrent scans to run. And that isn't ideal.

Although I suppose this sticks with the pre-existing functionality. So maybe that's not the worst thing, and we re-visit the concurrency model/safety at a later date.

tartufo/scanner.py
```diff
@@ -421,23 +436,24 @@ def evaluate_entropy_string(
...
         if not self.signature_is_excluded(string, chunk.file_path):
             entropy_score = self.calculate_entropy(string, chars)
             if entropy_score > min_entropy_score:
                 if self.entropy_string_is_excluded(string, line, chunk.file_path):
                     self.logger.debug("line containing entropy was excluded: %s", line)
                 else:
-                    return [Issue(types.IssueType.Entropy, string, chunk)]
-        return []
+                    issue = Issue(types.IssueType.Entropy, string, chunk)
+                    self._issues.append(issue)
```
tarkatronic: Let's try to keep this append at a higher level if we can. It should be possible to yield here, and maintain the generator functionality, but still only append the issues once at a higher level. That way there are no side effects on these lower-level functions.

tartufo/scanner.py
```diff
@@ -446,8 +462,8 @@ def scan_regex(self, chunk: types.Chunk) -> List[Issue]:
                     if not self.signature_is_excluded(match, chunk.file_path):
                         issue = Issue(types.IssueType.RegEx, match, chunk)
                         issue.issue_detail = key
-                        issues.append(issue)
-        return issues
+                        self._issues.append(issue)
```
tarkatronic: Same on this append here.

tests/test_cli.py
```diff
@@ -202,6 +202,7 @@ def test_command_exits_with_zero_return_code_when_no_issues_are_found(
         self, mock_scanner: mock.MagicMock
     ):
         mock_scanner.return_value.issues = []
+        mock_scanner.return_value.issue_count = 0
```
tarkatronic: Pretty sure this attribute went away 😄

tarkatronic (changes requested):
Some of this is probably just a bit of pontification (specifically the thread safety stuff), but I do want to be sure we don't add more side-effects where they did not previously exist.

tartufo/scanner.py
rbailey-godaddy: Only 1 scan runs. Suppose we will (eventually) return issues 1, 2, 3, ... 100.

Your scanner thread begins consuming `scanner.scan()` issues...

If your main thread starts the while loop *AFTER* the scanner thread finishes iterating:

* The scanner thread sees issues 1 - 100 in order
* The main thread exits the loop immediately

If your main thread starts the while loop *DURING* the scanner thread iteration (say just after issue 50 is viewed):

* The scanner thread sees issues 1-50 in order (before the main thread stuck its oar in the water), and then sees nothing more (because all subsequent issues were drained by `.issues`) and then, if it were called again, would return issues 1-100 in order (obtained by replaying `._issues`)...
* The main thread exits the loop immediately (after a delay during which issues 51-100 are enumerated)

If your main thread starts the while loop *BEFORE* the scanner thread starts iterating:

* The scanner thread sees issues 1-100 in order (returned by replaying `._issues`)
* The main thread exits the loop immediately (after a longer delay during which issues 1-100 are enumerated)

The only multithread wackiness would be in the fine print of the "DURING" case. Both are yielding issues from the low-level generators, which append to `._issues` just prior to yielding individual issues. So the question is, are generators and list append operations intrinsically thread-safe?

tests/test_cli.py
rbailey-godaddy: It did. I removed this one and a few of its buddies in other tests too.

tartufo/scanner.py
rbailey-godaddy: This (and the next) annoy me to hoist because I need to introduce loops inside `scan()`, which seems less clean to me, but I understand the issue of reducing side-effects. A commit with this change has been added.

tartufo/scanner.py
rbailey-godaddy: Ditto.

tartufo/scanner.py
rbailey-godaddy: The latest commit uses explicit locking to ensure all `scan()` invocations execute sequentially.

tartufo/scanner.py
tarkatronic: FWIW I agree, it's a tough balance and either way there is compromise to be made. I just tend to err on the side of reducing/condensing side effects in the hopes of maintaining a higher level of determinism and idempotency.

Thanks for updating this!

tartufo/scanner.py
tarkatronic: >  So the question is, are generators and list append operations intrinsically thread-safe?

Probably not 😅 This was largely me thinking out loud, and like I said we'll probably need to revisit the concurrency model later once we have an actual real use case. For now though, I think this will make for a great effort at correctness!

tarkatronic (approved):
Thanks for putting in all the hard work on this @rbailey-godaddy!

sushantmimani (approved):
LGTM 🦅

PR #222, opened 2021-09-16
allow exclude-entropy-patterns to match lines containing partial matches

kbartholomew-godaddy:
To help us get this pull request reviewed and merged quickly, please be sure to include the following items:

* [X] Tests (if applicable)
* [ ] Documentation (if applicable)
* [ ] Changelog entry
* [X] A full explanation here in the PR description of the work done

## PR Type
What kind of change does this PR introduce?

* [X] Bugfix
* [ ] Feature
* [ ] Code style update (formatting, local variables)
* [ ] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation content changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

Is this change backward compatible with the most recently released version? Does it introduce changes which might change the user experience in any way? Does it alter the API in any way?

* [X] Yes (backward compatible)
* [ ] No (breaking changes)

## Issue Linking

fixes #217 

## What's new?
- `ScannerBase.evaluate_entropy_string` now evaluates the entire line and uses `re.search()` when determining if an entropy string is included, so that false positives can be excluded based on their context.

   For example, `--exclude-entropy-patterns 'file.txt::falsePositive: [a-f0-9]{40}'` will exclude line 1 of this file but not line 2:

   ```yaml
   falsePositive: abcdef0123456789abcdef0123456789abcdef01
   realSecret: abcdef0123456789abcdef0123456789abcdef01
   ```

PR #229, opened 2021-10-14
New style entropy exclusions

tarkatronic:

* [X] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [X] Feature
* [X] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [X] Documentation content changes
* [X] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
- Reverted the behavior of `--exclude-entropy-patterns` to behave as it did before #222 by default
- Moved the new behavior introduced by `--exclude-entropy-patterns` behind an updated config specification style
- Updated our method of showing `DeprecationWarning`s

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -294,6 +294,42 @@ To exclude this, add ``docs/.*\.md$::^[a-zA-Z0-9]{40}$`` to ``exclude-entropy-pa
       "docs/.*\.md$::^[a-zA-Z0-9]{40}$", # exclude all git SHAs in the docs directory
     ]
 
+.. warning::
+    .. versionchanged:: 2.9.0
+    As of version 2.9.0, the above specification style has been deprecated, and
+    will be removed in version 3.0. The new style uses a TOML `array of tables`_
+    as shown below.
+
+    Note that this new syntax is not available on the command line, and must be
+    specified in your config file.
+
+Here is an example of how you might exclude SHA hashes in your docs, as well as
+hashes for GitHub Actions in your workflows:
+
+.. code-block:: toml
+
+    [tool.tartufo]
+    exclude-entropy-patterns = [
+        {files = 'docs/.*\.md$', match = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
+        {files = '\.github/workflows/.*\.yml', match = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
+    ]
+
+Thanks to the magic of TOML, you could also split these out into their own tables
+in the config if you wanted. So the following would be 100% equivalent to what
+you see above:
+
+.. code-block:: toml
+
+    [[tool.tartufo.exclude-entropy-patterns]]
+    files = 'docs/.*\.md$'
+    match = '^[a-zA-Z0-9]$'
+    reason = 'exclude all git SHAs in the docs'
+
+    [[tool.tartufo.exclude-entropy-patterns]]
+    files = '\.github/workflows/.*\.yml'
+    match = 'uses: .*@[a-zA-Z0-9]{40}'
+    reason = 'GitHub Actions'
+
 Limiting by Signature
 +++++++++++++++++++++
 
@@ -385,6 +421,7 @@ a devops pipeline.
 
 :doc:`examplecleanup`
 
+.. _array of tables: https://toml.io/en/v1.0.0#array-of-tables
 .. _install pre-commit: https://pre-commit.com/#install
 .. _pre-commit: https://pre-commit.com/
 .. _Shannon entropy: https://en.wiktionary.org/wiki/Shannon_entropy
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -310,8 +310,8 @@ hashes for GitHub Actions in your workflows:
 
     [tool.tartufo]
     exclude-entropy-patterns = [
-        {files = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
-        {files = '\.github/workflows/.*\.yml', pattern = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
+        {path-pattern = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
+        {path-pattern = '\.github/workflows/.*\.yml', pattern = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
     ]
 
 Thanks to the magic of TOML, you could also split these out into their own tables
@@ -321,12 +321,12 @@ you see above:
 .. code-block:: toml
 
     [[tool.tartufo.exclude-entropy-patterns]]
-    files = 'docs/.*\.md$'
+    path-pattern = 'docs/.*\.md$'
     pattern = '^[a-zA-Z0-9]$'
     reason = 'exclude all git SHAs in the docs'
 
     [[tool.tartufo.exclude-entropy-patterns]]
-    files = '\.github/workflows/.*\.yml'
+    path-pattern = '\.github/workflows/.*\.yml'
     pattern = 'uses: .*@[a-zA-Z0-9]{40}'
     reason = 'GitHub Actions'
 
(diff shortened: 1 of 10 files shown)


tartufo/scanner.py
```diff
@@ -295,13 +296,20 @@ def rule_matches(rule: Rule, string: str, path: str) -> bool:
...
-        if rule.path_pattern:
-            match = match and rule.path_pattern.search(path) is not None
+        if rule.re_match_type == "match":
+            if rule.pattern:
+                match = rule.pattern.match(string) is not None
+            if rule.path_pattern:
+                match = match and rule.path_pattern.match(path) is not None
+        elif rule.re_match_type == "search":
+            if rule.pattern:
+                match = rule.pattern.search(line) is not None
```
dclayton-godaddy, line 306: It seems like we're mixing two terms. Match type should be search vs match. However match_type seems to be changing the match scope from string to line. Should we provide match_scope (entropy vs line)?

docs/source/features.rst
```diff
@@ -294,6 +294,42 @@ To exclude this, add ``docs/.*\.md$::^[a-zA-Z0-9]{40}$`` to ``exclude-entropy-pa
...
+    Note that this new syntax is not available on the command line, and must be
+    specified in your config file.
+
+Here is an example of how you might exclude SHA hashes in your docs, as well as
+hashes for GitHub Actions in your workflows:
+
+.. code-block:: toml
+
+    [tool.tartufo]
+    exclude-entropy-patterns = [
```
dclayton-godaddy, line 312: Love this!

docs/source/features.rst
```diff
@@ -294,6 +294,42 @@ To exclude this, add ``docs/.*\.md$::^[a-zA-Z0-9]{40}$`` to ``exclude-entropy-pa
...
+        {files = 'docs/.*\.md$', match = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
```
dclayton-godaddy: Should we be using the term pattern where it's used in other places?

tartufo/scanner.py
tarkatronic: Yeah... I agree the naming is crap here. But this was my third iteration, and this split logic will only exist until v3.0, so I gave up on better naming. I'd be open to suggestions if you think it really is that bad.

docs/source/features.rst
tarkatronic: Good callout! I can make that change. That also reminds me that I forgot to add to the documentation, that's the only field that's absolutely required!

tartufo/scanner.py
dclayton-godaddy: If someone uses `search`, this change will search by line. Will this introduce a breaking change for 3.0? As in, will 3.0 no longer use `search` to search by line? If so and it's easy I'd recommend we add something like `match-scope` to avoid a breaking change.

docs/source/features.rst
```diff
@@ -294,6 +294,46 @@ To exclude this, add ``docs/.*\.md$::^[a-zA-Z0-9]{40}$`` to ``exclude-entropy-pa
...
+    ]
+
+Thanks to the magic of TOML, you could also split these out into their own tables
+in the config if you wanted. So the following would be 100% equivalent to what
+you see above:
+
+.. code-block:: toml
+
+    [[tool.tartufo.exclude-entropy-patterns]]
+    files = 'docs/.*\.md$'
```
dclayton-godaddy: should this be `file-pattern`?

dclayton-godaddy: or `path-pattern`?

tartufo/config.py
```diff
@@ -220,10 +221,14 @@ def load_rules_from_file(rules_file: TextIO) -> Dict[str, Rule]:
...
                 path_pattern=re.compile(path_pattern) if path_pattern else None,
+                re_match_type="match",
             )
         except AttributeError:
             rule = Rule(
-                name=rule_name, pattern=re.compile(rule_definition), path_pattern=None
+                name=rule_name,
+                pattern=re.compile(rule_definition),
+                path_pattern=None,
+                re_match_type="match",
```
rbailey-godaddy, line 231: I think your life would become much simpler if you used `re.compile("")` instead of `None` for the case that `path_pattern` is not specified. The empty regex matches immediately, which is what you're looking for, and the exact result of the match test is never used.

tartufo/config.py
```diff
@@ -255,20 +260,47 @@ def compile_rule(pattern: str) -> Rule:
         path, pattern = pattern.split("::", 1)
     except ValueError:  # Raised when the split separator is not found
         path = ".*"
```
rbailey-godaddy, line 262: Same here. I would use `""` instead of `".*"`, as it (also) matches every target every time, but doesn't require scanning the target string.

tartufo/config.py
```diff
@@ -255,20 +260,47 @@ def compile_rule(pattern: str) -> Rule:
...
-        for pattern in stripped
-        if pattern and not pattern.startswith("#")
-    ]
+    try:
+        return list(
+            {
+                Rule(
+                    name=pattern.get("reason", None),  # type: ignore[union-attr]
+                    pattern=re.compile(pattern["pattern"]),  # type: ignore[index]
+                    path_pattern=re.compile(pattern.get("files", ".*")),  # type: ignore[union-attr]
```
rbailey-godaddy:
```suggestion
                    path_pattern=re.compile(pattern.get("files", "")),  # type: ignore[union-attr]
```
For the same reasons as above

tartufo/scanner.py
```diff
@@ -295,13 +296,20 @@ def rule_matches(rule: Rule, string: str, path: str) -> bool:
...
         match = False
-        if rule.pattern:
-            match = rule.pattern.search(string) is not None
-        if rule.path_pattern:
-            match = match and rule.path_pattern.search(path) is not None
+        if rule.re_match_type == "match":
+            if rule.pattern:
+                match = rule.pattern.match(string) is not None
+            if rule.path_pattern:
+                match = match and rule.path_pattern.match(path) is not None
```
rbailey-godaddy, line 303: Because `rule.pattern` is mandatory and `rule.path_pattern` is always a `re` (if you took the advice above), I think this can be simplified:
```suggestion
            match = rule.pattern.match(string) is not None and rule.path_pattern.match(path) is not None
```
or presumably even `match = bool(rule.pattern.match(string) and rule.path_pattern.match(path))`

tartufo/scanner.py
rbailey-godaddy: +1 to this. It seems like it would be easy enough to use caller-supplied input to drive `.re_match_type`, and default to the behavior you have now. The benefit of doing it now rather than waiting for 3.0 is that users who actually pay attention to the deprecation warning could rewrite their configs in a way that would preserve behavior while making it explicit what they intended.

rbailey-godaddy: In fact, at the risk of making you tear out your hair, it might be better to allow (making this up on the fly) `scope: {line|string}` and transfer this directly to `.re_match_type` (or `.re_scope` if you prefer), where "match" becomes "string" and "search" becomes "line". Then we are aligned more with the intent vs the actual implementation, and it's ever so slightly easier to remember why we're doing this. ;-)

rbailey-godaddy (approved):
I made a few gratuitous suggestions, but the current code isn't bad. I'd lobby for the tweaks if the time is available, but if we're trying to get a quick fix out the door, go for it.

tartufo/config.py
tarkatronic: Given that this part is largely pre-existing code/behavior, I will make note of this for when we go back over this in preparation for 3.0

tartufo/scanner.py
tarkatronic: I think this can be handled in a follow-up easily enough, if it's desired badly enough. As-is, with the existing deprecation warning, users will be able to update their configs for the upcoming behavioral change.

docs/source/features.rst
tarkatronic: Hmmmm yeah I think maybe `file-pattern` would be good, since we're using it for matching file names.

tarkatronic: Although now that I look at the code, it's already referred to as `path_pattern`, so that seems to be the natural winner. 😄

smimani-godaddy (approved):
LGTM 🦅

sushantmimani (approved):

mayuriesha (approved):
LGTM

rbailey-godaddy (approved):
Still LGTM ;)

Issue #203, opened 2021-07-23
tartufo crashes scanning local repo with no remote named origin


jakeswenson:
## 🐛 Bug Report

When running `tartufo scan-local-repo .` it crashes since I don't have a remote named `origin` (I don't use standard remote names for my repositories.)

```bash
❯ tartufo scan-local-repo .
Traceback (most recent call last):
  File "/usr/local/bin/tartufo", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/tartufo/commands/scan_local_repo.py", line 47, in main
    scanner.scan()
  File "/usr/local/lib/python3.9/site-packages/tartufo/scanner.py", line 342, in scan
    for chunk in self.chunks:
  File "/usr/local/lib/python3.9/site-packages/tartufo/scanner.py", line 592, in chunks
    self._repo.remotes.origin.fetch()
  File "/usr/local/lib/python3.9/site-packages/git/util.py", line 972, in __getattr__
    return list.__getattribute__(self, attr)
AttributeError: 'IterableList' object has no attribute 'origin'
```

## To Reproduce

1. clone a git repository and `cd` into the repo directory
2. rename the origin remote. i.e. `git remote rename origin foo`
3. run `tartufo scan-local-repo .`
4. observe error above ^

## Expected Behavior

1. `tartufo` **should not fail** regardless of what the repo's remote names are. 
2. Why does `scan-local-repo` inspect remotes at all?

[...]

PR #248, opened 2021-10-29
Fix remote branch

### LINKED ISSUE #247: Specifying branch with remote repository is broken
## 🐛 Bug Report

`tartufo scan-remote-repo --branch <branch> <repo-uri>` No longer works.

In fact, some research indicates it hasn't worked since v2.0.2. Versions from v2.2.0 through v2.7.0 silently fail to discover secrets in the specified branch (I don't know what they are doing). Versions beginning with v2.8.0 claim the branch cannot be found, even though it exists.

rbailey-godaddy:

* [ ] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
fixes #247 

## What's new?
- Silently prefix user-supplied branch name with remote source name (i.e. "origin/") when user specifies `--branch` option with `scan-remote-repo`. The source name typically will be "origin", but it's looked up when the repository is cloned and we'll use whatever ends up being defined.

### DIFF
--- a/tartufo/commands/scan_remote_repo.py
+++ b/tartufo/commands/scan_remote_repo.py
@@ -56,7 +56,7 @@ def main(
     git_options = types.GitOptions(
         since_commit=since_commit,
         max_depth=max_depth,
-        branch=branch,
+        branch=f"origin/{branch}",
         include_submodules=include_submodules,
     )
     repo_path: Optional[Path] = None
--- a/tartufo/commands/scan_remote_repo.py
+++ b/tartufo/commands/scan_remote_repo.py
@@ -53,11 +53,10 @@ def main(
     include_submodules: bool,
 ) -> GitRepoScanner:
     """Automatically clone and scan a remote git repository."""
-    target_branch = f"origin/{branch}" if branch else None
     git_options = types.GitOptions(
         since_commit=since_commit,
         max_depth=max_depth,
-        branch=target_branch,
+        branch=None,
         include_submodules=include_submodules,
     )
     repo_path: Optional[Path] = None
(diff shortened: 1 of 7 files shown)


rbailey-godaddy:
FWIW, I did a quick check (manually specifying `--branch origin/<branch>`) against v2.9 (`main`) -- this fix doesn't work, so if we wanted to backport we'd need to do some more clever (or at least with more involved research).

sushantmimani (approved):
LGTM 🦅

tartufo/commands/scan_remote_repo.py
```diff
@@ -53,10 +53,11 @@ def main(
     include_submodules: bool,
 ) -> GitRepoScanner:
     """Automatically clone and scan a remote git repository."""
+    target_branch = f"origin/{branch}" if branch else None
```
tarkatronic: This might potentially re-introduce #203. Since we control the clone, can we be sure to specify the remote name as `origin` somehow? Or instead of hard-coding `origin` here, can we inspect the repository extract the remote name?

rbailey-godaddy: Our clone process always names the remote source "origin" (I checked). If the underlying library changes, that could be a problem. I'll look at putting in a step to retrieve the remote name just in case.

tarkatronic: My concern is, `GitPython` uses actual `git` commands under the surface, which ends up reading user config... so could a user have their config set up such that it automatically clones with a different name?

rbailey-godaddy: I just added a commit that returns the actual remote source when the repository is cloned, and then uses that.

tarkatronic (approved):

PR #233, opened 2021-10-19
Switch to Pygit2

sushantmimani:
To help us get this pull request reviewed and merged quickly, please be sure to include the following items:

* [x] Tests (if applicable)
* [ ] Documentation (if applicable)
* [ ] Changelog entry
* [x] A full explanation here in the PR description of the work done

## PR Type
What kind of change does this PR introduce?

* [ ] Bugfix
* [ ] Feature
* [ ] Code style update (formatting, local variables)
* [x] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Documentation content changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

Is this change backward compatible with the most recently released version? Does it introduce changes which might change the user experience in any way? Does it alter the API in any way?

* [ ] Yes (backward compatible)
* [x] No (breaking changes)
In its current version, this PR will flag findings for renamed files even if the findings in the new file(after rename) have been excluded from Tartufo.

## What's new?
This is the first PR in a series of PRs aimed at re-implementing #122. Current status:
- scan-remote-repo runs successfully
- scan-local-repo runs successfully
- pre-commit runs successfully
- All unit tests pass
- `--max-depth` option is unimplemented
- file renames are not tracked so extra findings ( which might have exclusions for the new file path) are generated

PR #249, opened 2021-10-29
Handling file renames with pygit2

mayuriesha:

* [x] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [ ] Feature
* [x] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [x] No (breaking changes)

## What's new?
- This is a continuation of migration from GitPython to Pygit2 to handle file renames gracefully [PR-233](https://github.com/godaddy/tartufo/pull/233)
- All unit tests pass

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -542,30 +542,23 @@ class GitScanner(ScannerBase, abc.ABC):
             if delta.is_binary:
                 self.logger.debug("Binary file skipped: %s", file_path)
                 continue
-            delta_patch_status = delta.status
-            if delta_patch_status == pygit2.GIT_DELTA_DELETED:
+            if delta.status == pygit2.GIT_DELTA_DELETED:
                 self.logger.debug("Skipping as the file is deleted")
                 continue
             printable_diff: str = patch.text
-            lines_to_truncate = self.header_line_count(printable_diff)
-            printable_diff = (
-                printable_diff.split("\n", lines_to_truncate)[lines_to_truncate]
-                if lines_to_truncate > 0
-                else ""
-            )
+            header_length = GitScanner.header_length(printable_diff)
+            printable_diff = printable_diff[header_length:]
             if self.should_scan(file_path):
                 # The `printable_diff` contains the full 4-line diff header,
                 # so we need to strip that before analyzing it
                 yield printable_diff, file_path
 
-    def header_line_count(self, diff: str) -> int:
-        """Computes the git diff header length"""
-        lines_to_remove: int = 0
-        for line_no, data in enumerate(diff.split("\n")):
-            if data.startswith("+++"):
-                lines_to_remove = line_no + 1
-                break
-        return lines_to_remove
+    @staticmethod
+    def header_length(diff: str) -> int:
+        """Compute the length of the git diff header text"""
+        b_file_pos = diff.find("+++")
+        final_nl = diff.find("\n", b_file_pos)
+        return final_nl + 1
 
     def filter_submodules(self, repo: pygit2.Repository) -> None:
         """Exclude all git submodules and their contents from being scanned."""
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -561,7 +561,7 @@ class GitScanner(ScannerBase, abc.ABC):
             b_file_pos = diff.index("\n+++")
             return diff.index("\n", b_file_pos + 4) + 1
         except ValueError:
-            # If no separator, diff is header only (pure rename)
+            # Diff is pure header as it is a pure rename(similarity index 100%)
             return len(diff)
 
     def filter_submodules(self, repo: pygit2.Repository) -> None:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -534,8 +534,6 @@ class GitScanner(ScannerBase, abc.ABC):
 
         :param diff_index: The diff index / commit to be iterated over
         """
-
-        diff.find_similar(rename_threshold=100)
         for patch in diff:
             delta: pygit2.DiffDelta = patch.delta
             file_path = (
@@ -559,7 +557,7 @@ class GitScanner(ScannerBase, abc.ABC):
                 yield printable_diff.split("\n", lines_to_truncate)[lines_to_truncate], file_path
 
     def update_printable_diff(self, diff: str) -> int:
-        """Removes the unwanted content from printable diff"""
+        """Computes the git diff header length"""
         lines_to_remove: int = 0
         for line_no, data in enumerate(diff.split("\n")):
             if data.startswith("+++"):
@@ -715,6 +713,7 @@ class GitRepoScanner(GitScanner):
                 if diff_hash in already_searched:
                     continue
                 already_searched.add(diff_hash)
+                diff.find_similar(rename_threshold=100)
                 for blob, file_path in self._iter_diff_index(diff):
                     yield types.Chunk(
                         blob,
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -543,18 +543,20 @@ class GitScanner(ScannerBase, abc.ABC):
                 self.logger.debug("Binary file skipped: %s", file_path)
                 continue
             delta_patch_status = delta.status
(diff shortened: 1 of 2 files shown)


rbailey-godaddy (changes requested):
So there are a few things that trouble me about the extraction of the body we want to scan from printable_diff.

First, you are splitting the text (potentially expensive, especially if the diff is large) twice -- once in `header_line_count()` and a second time in `_iter_diff_index()` itself when you use the results from the first function.

It would be more efficient to do the split once, and then -- since you know where the payload is -- simply return the payload instead of making the caller duplicate most of your work to obtain it.

Second, both the old and new logic assume the payload is "one line" (one array element), so I presume any newlines in the payload are escaped. It occurred to me to wonder if the payload is always the last element of `.text`, in which case it would be more straightforward to use something like `printable_diff.rsplit("\n", 1)[-1]` (i.e. working backwards from the end, find the last newline, split only there, and return everything following it.)

tartufo/scanner.py
```diff
@@ -543,11 +542,30 @@ def _iter_diff_index(
...
+                yield printable_diff, file_path
+
+    def header_line_count(self, diff: str) -> int:
+        """Computes the git diff header length"""
+        lines_to_remove: int = 0
+        for line_no, data in enumerate(diff.split("\n")):
+            if data.startswith("+++"):
+                lines_to_remove = line_no + 1
+                break
+        return lines_to_remove
```
tarkatronic: I agree with @rbailey-godaddy that this is a needlessly expensive operation, especially when we already have concerns about memory usage. I wonder if a different approach would be better. Instead of `header_line_count`, what if we do `header_length` and have something like:

```python
@staticmethod
def header_length(diff: str) -> int:
    """Compute the length of the git diff header text"""
    b_file_pos = diff.find("+++")
    final_nl = diff.find("\n", b_file_pos)
    return final_nl + 1
```

Then we can use it something like...

```python
header_length = GitScanner.header_length(diff)
printable_diff = printable_diff[header_length:]
```

sushantmimani: Second this approach.

tartufo/scanner.py
```diff
@@ -543,11 +542,30 @@ def _iter_diff_index(
             if delta.is_binary:
                 self.logger.debug("Binary file skipped: %s", file_path)
                 continue
+            delta_patch_status = delta.status
+            if delta_patch_status == pygit2.GIT_DELTA_DELETED:
```
sushantmimani:
```suggestion
            if delta.status == pygit2.GIT_DELTA_DELETED:
```

tartufo/scanner.py
```diff
@@ -543,11 +542,23 @@ def _iter_diff_index(
...
             if self.should_scan(file_path):
                 # The `printable_diff` contains the full 4-line diff header,
                 # so we need to strip that before analyzing it
-                yield printable_diff.split("\n", 4)[4], file_path
+                yield printable_diff, file_path
+
+    @staticmethod
+    def header_length(diff: str) -> int:
+        """Compute the length of the git diff header text"""
+        b_file_pos = diff.find("+++")
```
rbailey-godaddy: Color me paranoid, but I think we want this:
```suggestion
        b_file_pos = diff.find("\n+++") + 4
```

rbailey-godaddy (changes requested):
One more corner case - so we don't go spastic where +++ appears at random places in the diff metadata.

tartufo/scanner.py
tarkatronic: I understand (and agree with) the `\n`, but why `+ 4`?

rbailey-godaddy: Well, if we do nothing, then `b_file_pos` points at the first character of the find, namely `\n` (instead of the first `+` as previously). Therefore the next find will match immediately, and the current position (`+++` followed by whatever) will be returned as the first part of the diff payload. In actuality, we know the newline we're seeking can't come any sooner than immediately following `+++`, i.e. 4 characters after the `\n` we matched.

tarkatronic: Got it. Okay, that makes sense!

tarkatronic (approved):

tartufo/scanner.py
sushantmimani: I think we need to add a comment here to explain the rationale (what @rbailey-godaddy said above) for using the `+ 4`

mayuriesha: This might work differently in a specific scenario where the file is a rename with similarity index being 100%. The printable diff in this case would look something as follows:

```
diff --git a/.github/actions/wss-unified-agent.config b/.github/actions/run-wss-scan/wss-unified-agent.config
similarity index 100%
rename from .github/actions/wss-unified-agent.config
rename to .github/actions/run-wss-scan/wss-unified-agent.config
```

In this case there will not be a `+++`, so it will end up truncating the first line in the above diff. I believe it should be okay. Just wanted to get it to your notice before I make changes

tarkatronic: Interesting, then maybe we have something like

```suggestion
        try:
            b_file_pos = diff.find("\n+++") + 4  # First character after the +++
            final_nl = diff.find("\n", b_file_pos)
        except ValueError:
            final_nl = -1  # This will return 0, since we add 1 at the end
```

mayuriesha: It will not raise an error. Instead diff.find("\n+++") will return -1. So going with the following code snippet if it looks okay?

```
    @staticmethod
    def header_length(diff: str) -> int:
        """Compute the length of the git diff header text"""
        b_file_pos = diff.find("\n+++")
        if b_file_pos != -1:
            b_file_pos = b_file_pos + 4  # Retrieving first character after the +++
        final_nl = diff.find("\n", b_file_pos)
        return final_nl + 1
```
Retrieving exact position if there is an occurrence of "\n+++"

tarkatronic: Ah, right! I was getting confused between `.find()` and `.index()`. Using `.find()` returns `-1`, where `.index()` will raise a `ValueError` 😄 
Looks great!

tartufo/scanner.py
```diff
@@ -543,11 +542,23 @@ def _iter_diff_index(
...
                 continue
+            if delta.status == pygit2.GIT_DELTA_DELETED:
+                self.logger.debug("Skipping as the file is deleted")
+                continue
             printable_diff: str = patch.text
+            header_length = GitScanner.header_length(printable_diff)
+            printable_diff = printable_diff[header_length:]
             if self.should_scan(file_path):
                 # The `printable_diff` contains the full 4-line diff header,
                 # so we need to strip that before analyzing it
```
sushantmimani: These comments need to be moved up

tartufo/scanner.py
rbailey-godaddy: I still can't follow this. What about:

```
@staticmethod
def header_length(diff: str) -> int:
    """Compute the length of the git diff header text"""
    try:
        # Header ends after newline following line starting with "+++"
        b_file_pos = diff.index("\n+++")
        return diff.index("\n", b_file_pos + 4) + 1
    except ValueError:
        # If no separator, diff is header only (pure rename)
        return len(diff)
```

mayuriesha: Sure this looks good and we can have this. 
By returning len(diff) in except block we are not going to process the header at all in case of pure renames.

tartufo/scanner.py
```diff
@@ -543,11 +542,27 @@ def _iter_diff_index(
             if delta.is_binary:
                 self.logger.debug("Binary file skipped: %s", file_path)
                 continue
+            if delta.status == pygit2.GIT_DELTA_DELETED:
+                self.logger.debug("Skipping as the file is deleted")
+                continue
             printable_diff: str = patch.text
+            # The `printable_diff` contains the full 4-line diff header,
```
sushantmimani: The header might be 4 or 5 lines so the comment must be updated accordingly .

rbailey-godaddy (approved):
LGTM

sushantmimani (approved):
LGTM 🦅

PR #250, opened 2021-10-30
Fix scan-remote-repo --branch logic

### LINKED ISSUE #247: Specifying branch with remote repository is broken

rbailey-godaddy:

* [X] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
fixes #247

## What's new?
- A fresh-cloned repository does not have branch information for remote branches. We must follow the information in `remotes.origin.refs` instead of `branches` (which will have only the main/master branch).
- We add a new `is_remote` flag so the common code knows which strategy to employ.
- A side-effect appears to be that because `HEAD` appears in the reference list, we'll scan the default branch twice, but I can live with this.

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -632,24 +632,47 @@ class GitRepoScanner(GitScanner):
         already_searched: Set[bytes] = set()
 
         try:
-            if self.git_options.branch:
-                # Single branch only
-                if self.git_options.fetch:
-                    self._repo.remotes.origin.fetch(self.git_options.branch)
-                unfiltered_branches = list(self._repo.branches)
-                branches = [
-                    x for x in unfiltered_branches if x.name == self.git_options.branch
-                ]
+            if self.git_options.is_remote:
+                # Remote repository - local directory does not have branch info
+                # and we must follow remote references
+                unfiltered_branches = list(self._repo.remotes.origin.refs)
+                if self.git_options.branch:
+                    # Single branch only
+                    branches = [
+                        x
+                        for x in unfiltered_branches
+                        if x.name == f"origin/{self.git_options.branch}"
+                    ]
 
-                if len(branches) == 0:
-                    raise BranchNotFoundException(
-                        f"Branch {self.git_options.branch} was not found."
-                    )
+                    if len(branches) == 0:
+                        raise BranchNotFoundException(
+                            f"Branch {self.git_options.branch} was not found."
+                        )
+                else:
+                    # Everything
+                    branches = unfiltered_branches
             else:
-                # Everything
-                if self.git_options.fetch:
-                    self._repo.remotes.origin.fetch()
-                branches = list(self._repo.branches)
+                # Local repository
+                if self.git_options.branch:
+                    # Single branch only
+                    if self.git_options.fetch:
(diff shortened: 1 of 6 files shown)


tartufo/scanner.py
```diff
@@ -632,24 +632,47 @@ def chunks(self) -> Generator[types.Chunk, None, None]:
...
+                        if x.name == f"origin/{self.git_options.branch}"
+                    ]
+
+                    if len(branches) == 0:
+                        raise BranchNotFoundException(
+                            f"Branch {self.git_options.branch} was not found."
+                        )
+                else:
+                    # Everything
+                    branches = unfiltered_branches
```
tarkatronic, line 653: Will this truly only be branches, or might this possibly include refs to PRs, tags, etc?

tarkatronic (commented):
Just one question/point of curiosity, to be sure we're not being excessive in our scan.

Also, is this something we're going to need to port over to the v3.x branch?

tarkatronic:
(Ignore that last part of my comment -- just saw the other PR!)

tartufo/scanner.py
rbailey-godaddy: It might be excessive (I know in simple testing that `HEAD` showed up in the references list). However, I feel this is "harmlessly excessive":
* If user specifies a branch, we are looking for a reference with the name of the branch, and therefore will scan only the branch the user asked us to scan
* If the user didn't, we'll scan "everything" (which is what is intended). We might end up scanning parts of "everything" multiple times, to the extent multiple references appear -- in which case I hope most of this will be short-circuited by the "I've seen this diff before" lru cache and the impact will be minimized.

The fix for v3.x is much better than this one, so I'm willing to sacrifice some efficiency for correctness in the short term, knowing that when people transition to v3 it won't be a problem any longer.

tarkatronic: Makes sense. 👍🏻

tarkatronic (approved):

tartufo/scanner.py
rbailey-godaddy: Subtle change: In testing, I noted that the fresh-cloned scratch repo had only a main (or presumably master) branch defined, so the old default `scan-remote-repo` (without a branch) actually was probably only scanning the master branch and not the entire repository -- so people might be surprised to discover things getting scanned now that weren't scanned (at least not since v2.0.2).

tarkatronic: I think it's likely there will be a number of surprises for users as they move to v3.0 and find how much more accurate it is. 😄

smimani-godaddy (approved):
LGTM 🦅

sushantmimani (approved):
LGTM 🦅

PR #258, opened 2021-11-02
Version bump for v2.9.1

mayuriesha:

* [ ] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-vx.y.z - TBD
+v2.9.1 - 3 November 2021
 ------------
 
 Bug fixes:
@@ -8,6 +8,14 @@ Bug fixes:
   v2.7.0 failed silently (not scanning the branch, and returning no error).
   Versions v2.8.0 and later claimed the branch did not exist, even if it did.
   This option now works correctly.
+
+Features:
+
+* [#231](https://github.com/godaddy/tartufo/issues/231) Change toml parsing library to use tomlkit
+
+Other changes:
+
+* [#251](https://github.com/godaddy/tartufo/issues/251) Document update to use --no-fetch flag to all scan-local-repo
 
 v2.9.0 - 19 October 2021
 ------------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ maintainers = ["GoDaddy <oss@godaddy.com>"]
 name = "tartufo"
 readme = "README.md"
 repository = "https://github.com/godaddy/tartufo/"
-version = "2.9.0"
+version = "2.9.1"
 
 [tool.poetry.scripts]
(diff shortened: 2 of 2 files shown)


tarkatronic (approved):

rbailey-godaddy (changes requested):
In hindsight, looking for @tarkatronic concurrence... I think I would make the wording ending in L9-10 stronger.

The problem we explicitly set about fixing was that you couldn't say `scan-remote-repo --branch foobar` because the `.branches` collection only had `main` and nothing else, so your branch could never find it. But I think that isn't what got exercised the most (since nobody complained) and doesn't represent the biggest exposure.

What we didn't think about so much is the case of `scan-remote-repo` (without a branch). This is supposed to scan all branches... but as I just mentioned, the only branch it knows about is `main` (or `master` I suppose), so that is the **only** branch it would scan. If you had other branches, and those had secrets, They. Were. Not. Scanned. I just tested this behavior with 2.9.0  (which had one kind of brain damage) and also with 2.2.0 (which had the different brain damage) and with no branch specified, you get no errors *AND* no non-main branches get scanned and you end up with a massive false sense of security. :(

rbailey-godaddy:
FYI the testing cited above was done using this:

```sh
python tartufo/__main__.py --regex --no-entropy scan-remote-repo git@ghc:gdcorp-appservices/tartufo-test.git
```

The `rename-1` branch in that repository has several regex-based secrets that should be discovered, but won't be found by most of the v2 releases.

tarkatronic:
> In hindsight, looking for @tarkatronic concurrence... I think I would make the wording ending in L9-10 stronger.
> 
> The problem we explicitly set about fixing was that you couldn't say `scan-remote-repo --branch foobar` because the `.branches` collection only had `main` and nothing else, so your branch could never find it. But I think that isn't what got exercised the most (since nobody complained) and doesn't represent the biggest exposure.
> 
> What we didn't think about so much is the case of `scan-remote-repo` (without a branch). This is supposed to scan all branches... but as I just mentioned, the only branch it knows about is `main` (or `master` I suppose), so that is the **only** branch it would scan. If you had other branches, and those had secrets, They. Were. Not. Scanned. I just tested this behavior with 2.9.0 (which had one kind of brain damage) and also with 2.2.0 (which had the different brain damage) and with no branch specified, you get no errors _AND_ no non-main branches get scanned and you end up with a massive false sense of security. :(

So... what is your actual proposed change here?

mayuriesha:
Closing this PR as we will be releasing v2.10.0 and this branch was named v2.9.1

PR #266, opened 2021-11-04
Dropping support for --json and --compact and adding support for --output-format

### LINKED ISSUE #253: Consolidate output options into `--output-format` / `-of` flag

During the discussion that resulted in https://github.com/godaddy/tartufo/discussions/239, we talked about the current output formats, the various options to trigger them, along with any possible future formats, and decided that it would be best to combine them into a single option. There really is only ever one applicable output format, making them somewhat mutually exclusive, so this will both improve the user experience and reduce complexity in the code.

The following options should be removed:

* `--json` / `--no-json`
* `--compact` / `--no-compact`

The following option should be added:

`--output-format [json | compact | text]` (Default: `text`)

This also makes for extensibility in the future in case we want, for example, `--output-format [toml | yaml]` or whatever else. This will also be easier to document and logic around in the code.

mayuriesha:

* [ ] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
closes #253

## What's new?
- The PR drops support for `--json` and `--compact` options and instead uses single flag `--output-format json/compact/text` to support different output formats

### DIFF
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -45,7 +45,6 @@ class TartufoCLI(click.MultiCommand):
     name="tartufo",
     context_settings=dict(help_option_names=["-h", "--help"]),
 )
-@click.option("--json/--no-json", help="Output in JSON format.", is_flag=True)
 @click.option(
     "--rules",
     multiple=True,
@@ -59,13 +58,6 @@ class TartufoCLI(click.MultiCommand):
     show_default=True,
     help="Whether to include the default regex list when configuring"
     " search patterns. Only applicable if --rules is also specified.",
-)
-@click.option(
-    "--compact/--no-compact",
-    is_flag=True,
-    default=False,
-    show_default=True,
-    help="Enable reduced output.",
 )
 @click.option(
     "--entropy/--no-entropy",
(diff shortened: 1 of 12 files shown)


tartufo/cli.py
```diff
@@ -122,6 +114,13 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
     are excluded unless effectively excluded via the --include-path-patterns
     option.""",
 )
+@click.option(
+    "-of",
+    "--output-format",
+    type=click.Choice(["json", "compact", "text"]),
+    default="text",
+    help="Output can be in json/compact/text format. By default the output will be in text format.",
```
tarkatronic: I really like the description you added to the `README.md` for this option -- can you put that in here?

tartufo/scanner.py
```diff
@@ -68,12 +68,12 @@ def __init__(
         self.chunk = chunk
         self.logger = logging.getLogger(__name__)
 
-    def as_dict(self, compact=False) -> Dict[str, Optional[str]]:
+    def as_dict(self, output_format=None) -> Dict[str, Optional[str]]:
```
tarkatronic: I think it makes the most sense to keep this method the way it was, using a boolean. Since the other values of `output_format` mean nothing to this, a boolean seems cleanest.

tarkatronic (changes requested):
Just a couple of small changes here. The more I see this, the more I love the decision to make this change!

One thing I'm wondering as I look through this code though -- does it make sense to make an enum? I always hate having to pass around magical string values where typoes could get introduced. What about a `types.OutputFormat`? I'm very much open to arguments either way on this. 😄

sushantmimani (approved):
LGTM, once @tarkatronic's comments have been addressed.

tarkatronic (approved):
Looks great!

PR #272, opened 2021-11-09
Add --entropy-sensitivity option for controlling entropy checks

### LINKED ISSUE #265: Change `--b64-entropy-score` and `--hex-entropy-score` to `--entropy-sensitivity`

The `--b64-entropy-score` and `--hex-entropy-score` arguments, while useful, can be very confusing. Even knowing the internals of tartufo like I do, I can't say with complete confidence what the default values represent, nor how they were arrived at.

As described by @rbailey-godaddy in https://github.com/godaddy/tartufo/pull/223#issuecomment-934404699, it would be far more useful and intuitive to have a `--entropy-sensitivity` argument, ranging from 1-100. It would require a bit more calculation on the part of tartufo, but would then be much more understandable for users.

The one thing I'm unsure of is, should this be split out into `--b64-entropy-sensitivity` and `--hex-entropy-sensitivity`, or keep them combined as one?

rbailey-godaddy:

* [X] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [ ] Bugfix
* [X] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
fixes #265

## What's new?

This PR introduces a new `--entropy-sensitivity` option that replaces `--b64-entropy-score` and `--hex-entropy-score`. The older options are marked "deprecated" and generate warnings when used.

The default sensitivity is exactly the same as it was before, but it is presented differently and applied consistently to both hexadecimal and base64 strings. "Sensitivity" is expressed on a scale of 0 (totally random) to 100 (totally nonrandom) -- the default is 25 -- and intuitively higher sensitivity means more borderline-random strings will be reported as suspicious.

For purposes of minimizing this PR's footprint, the default is applied both in `cli.py` (real world) and in `scanner.py` (because existing unit tests do not specify most options).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ Features:
   `--scan-filenames/--no-scan-filenames` flag which allows users to enable or disable file name scanning.
 * [#254](https://github.com/godaddy/tartufo/pull/260) - Changes the default value of
   `--regex/--no-regex` to True.
-* [#265](https://github.com/godaddy/tartufo/issues/265) - Adds new `--sensitivity`
+* [#265](https://github.com/godaddy/tartufo/issues/265) - Adds new `--entropy-sensitivity`
   option which provides a friendlier way to adjust entropy detection sensitivity.
   This replaces `--b64-entropy-score` and `--hex-entropy-score`, which now are
   marked as deprecated.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ Features:
   `--scan-filenames/--no-scan-filenames` flag which allows users to enable or disable file name scanning.
 * [#254](https://github.com/godaddy/tartufo/pull/260) - Changes the default value of
   `--regex/--no-regex` to True.
+* [#265](https://github.com/godaddy/tartufo/issues/265) - Adds new `--sensitivity`
+  option which provides a friendlier way to adjust entropy detection sensitivity.
+  This replaces `--b64-entropy-score` and `--hex-entropy-score`, which now are
+  marked as deprecated.
 
 Misc:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,15 @@
 vx.y.z - TBD
 ------------
 
+Features:
+
 * [#270](https://github.com/godaddy/tartufo/issues/270) - When no refs/branches
   are found locally, tartufo will now scan the repo HEAD as a single commit,
   effectively scanning the entire codebase at once.
+* [#265](https://github.com/godaddy/tartufo/issues/265) - Adds new `--entropy-sensitivity`
+  option which provides a friendlier way to adjust entropy detection sensitivity.
+  This replaces `--b64-entropy-score` and `--hex-entropy-score`, which now are
+  marked as deprecated.
 
 v3.0.0-alpha.1 - 11 November 2021
 ---------------------------------
@@ -27,10 +33,6 @@ Features:
   `--scan-filenames/--no-scan-filenames` flag which allows users to enable or disable file name scanning.
 * [#254](https://github.com/godaddy/tartufo/pull/260) - Changes the default value of
   `--regex/--no-regex` to True.
-* [#265](https://github.com/godaddy/tartufo/issues/265) - Adds new `--entropy-sensitivity`
-  option which provides a friendlier way to adjust entropy detection sensitivity.
-  This replaces `--b64-entropy-score` and `--hex-entropy-score`, which now are
-  marked as deprecated.
 
 Misc:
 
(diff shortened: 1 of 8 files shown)


sushantmimani (commented):
Does it make more sense to add deprecation warnings to v2.x for this and any other option(s) that we want to remove in v3.0 & then just remove them in v3.0?

rbailey-godaddy:
> Does it make more sense to add deprecation warnings to v2.x for this and any other option(s) that we want to remove in v3.0 & then just remove them in v3.0?

I dunno. This feels a lot like obeying the letter of the law ("we warned you in 2.11 (?) that this would disappear in 3.0") while violating the spirit by jamming it in at chronologically the last second and effectively saying "you don't HAVE to get all the new features in 3.0, so you can take as long as you want to convert, and suffer in the meantime...".

The extra logic/code for the backward compatibility is pretty minor and cleanly separated, so I don't view it as a problem to leave it for now (marked as deprecated) and then tear it out in 3.future or even 4.0. Especially if it means we can avoid yet another 2.x update right on the heels of all the ones we already published.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ Features:
   `--scan-filenames/--no-scan-filenames` flag which allows users to enable or disable file name scanning.
 * [#254](https://github.com/godaddy/tartufo/pull/260) - Changes the default value of 
   `--regex/--no-regex` to True.
+* [#265](https://github.com/godaddy/tartufo/issues/265) - Adds new `--sensitivity`
```
tarkatronic:
```suggestion
* [#265](https://github.com/godaddy/tartufo/issues/265) - Adds new `--entropy-sensitivity`
```

sushantmimani (approved):
LGTM 🦅

tartufo/scanner.py
```diff
@@ -142,11 +142,46 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
...
+        if self.global_options.hex_entropy_score:
+            warnings.warn(
+                "--hex-entropy-score is deprecated. Use --entropy-sensitivity instead."
+            )
+            self._hex_entropy_score = self.global_options.hex_entropy_score
+
+        if self.global_options.b64_entropy_score:
+            warnings.warn(
+                "--b64-entropy-score is deprecated. Use --entropy-sensitivity instead."
+            )
```
```suggestion
        if self.global_options.hex_entropy_score:
            warnings.warn(
                "--hex-entropy-score is deprecated. Use --entropy-sensitivity instead.",
                DeprecationWarning
            )
            self._hex_entropy_score = self.global_options.hex_entropy_score

        if self.global_options.b64_entropy_score:
            warnings.warn(
                "--b64-entropy-score is deprecated. Use --entropy-sensitivity instead.",
                DeprecationWarning
            )
```
It's important that these are specifically deprecation warnings, or the user will likely not see them: https://github.com/godaddy/tartufo/blob/3b4acef809158676ebf7dfa8b93ae6f0a525b199/tartufo/cli.py#L277-L278

tarkatronic: I think it's fine for these deprecation warnings to stay here; I think the new calculations should move to properties though, staying consistent with all the rest of our calculated instance properties. Or of course the warnings can move with them. Either way. But I think let's keep our calculated things in properties :)

tarkatronic (changes requested):
With the update to the command line options, please remember to update the `--help` output in the `README.md`!

CHANGELOG.md
rbailey-godaddy: Fixed. I missed this one after I forgot what was proposed and started off with the wrong option name. :-/

tartufo/scanner.py
rbailey-godaddy: Assuming you meant "expose these as properties instead of internal members", it's done now.

rbailey-godaddy: And the warnings are fixed -- thanks for the information, I was clueless regarding the distinction or handling!

tarkatronic: I actually meant having the calculations in the properties as well, keeping the init as minimal as possible, and all calculations as JIT 😄 Same as we do with all our exclusion patterns, etc.

tarkatronic: The other advantage of using properties is that it cuts down on the number of instance variables, potentially appeasing pylint's warning

rbailey-godaddy: I just added another commit which removes those two members at the cost of nearly doubling the related code. But pylint and mypy are pytest are happy, so it must be good. :)

rbailey-godaddy: ...and fixed again per this morning's review meeting.

tarkatronic (approved):
🎉

PR #274, opened 2021-11-09
Entropy scanning optimizations and base64url support

### LINKED ISSUE #177: tartufo entropy scanning does not handle base64url encoding
## 🐛 Bug Report

There are multiple flavors (distinct in the alphabet used) of base64.

Tartufo source code at `tartufo/scanner.py` has
```
BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
```

This doesn't match for example the base 64 "url" flavor that uses `_` and `-` (instead of `+`. and `/`) characters as defined in §5 of RFC4648.

## To Reproduce

```
$ git init test
$ cd test
$ git commit --allow-empty --allow-empty-message -m 'Start'
[master (root-commit) a64409a] Start
$ echo -e "g34\xFFahChi\xFAk91PWpbG" | base64 > secret_base64.txt
$ echo -e "g34\xFFahChi\xFAk91PWpbG" | base64 | tr + - | tr / _ > secret_base64url.txt
$ cat secret_base64*
ZzM0/2FoQ2hp+ms5MVBXcGJHCg==
ZzM0_2FoQ2hp-ms5MVBXcGJHCg==
$ git add secret_base64*
$ tartufo --entropy --regex pre-commit
~~~~~~~~~~~~~~~~~~~~~
Reason: High Entropy
Filepath: secret_base64.txt
Signature: 56d5080cddeeb2c434550afd809c43c06bef51f799b96cdc697f436e5270350e
@@ -0,0 +1 @@
+ZzM0/2FoQ2hp+ms5MVBXcGJHCg==

~~~~~~~~~~~~~~~~~~~~~
```

## Expected Behavior

All flavors of base64 should be treated the same concerning entropy, so in above example both files should be reported.

## Code Example

N/A

## Environment

tartufo version 2.4.0

### LINKED ISSUE #273: Optimize entropy measuring functions

Entropy support functions are unnecessarily complex and inefficient. Reimplement them to be more comprehensible.

rbailey-godaddy:

* [X] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [ ] Bugfix
* [X] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [X] Yes (backward compatible) -- with minor/rare corner cases noted below
* [ ] No (breaking changes)

## Issue Linking

closes #177 
closes #273

## What's new?

Several key components have been rewritten for better performance and improved maintainability:

* `get_strings_of_set()` has been replaced with `find_string_encodings()` -- Previously we performed matching ourselves, a character at a time. I'm betting a standard regex will be at least as or more efficient. It's also more accurate; in particular the base64 pad character `=` now is allowed only at the end of an encoding, as specified by RFC4648.
* `calculate_entropy()` -- Previously we used the encoding alphabet (unnecessarily) as a loop range, and scanned the input data once for every character in the alphabet. Now we scan the input once and use a Counter to generate probability data.
* `evaluate_entropy_string()` no longer requires the encoding alphabet as input
* `scan_entropy()` was reorganized to avoid buffering intermediate results
* `BASE64_CHARS` and `HEX_CHARS` have been replaced by `BASE64_REGEX` and `HEX_REGEX` respectively.
* Finally, `BASE64_REGEX` has been extended to include `-` and `_` so it discovers base64url strings in addition to base64 strings.

All unit tests pass after only changes to fix up call signatures. Possible regressions are these, none of which are expected to appear in real life:
* A string containing `=` in the middle would have been recognized as base64 but now it will be recognized as two base64 strings (the first ending with the `=` and the second containing the remainder of the string). `=` is not allowed to appear in a base64 (or base64url) string except at the end.
* A base64 string adjacent to either `-` or `_` will now be recognized as a longer string containing those characters. These are not valid base64 but are valid base64url and the scanner does not distinguish between the two (or a mutant combination of both of them).

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -30,7 +30,7 @@ import pygit2
 from tartufo import config, types, util
 from tartufo.types import BranchNotFoundException, Rule, TartufoException
 
-BASE64_REGEX = re.compile(r"[A-Z0-9+/]+={,2}", re.IGNORECASE)
+BASE64_REGEX = re.compile(r"[A-Z0-9+/_-]+={,2}", re.IGNORECASE)
 HEX_REGEX = re.compile(r"[0-9A-F]+", re.IGNORECASE)
 
 
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -469,11 +469,11 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
             for word in line.split():
                 for string in util.find_strings_by_regex(word, BASE64_REGEX):
                     yield from self.evaluate_entropy_string(
-                        chunk, line, string, BASE64_CHARS, self.b64_entropy_limit
+                        chunk, line, string, self.b64_entropy_limit
                     )
                 for string in util.find_strings_by_regex(word, HEX_REGEX):
                     yield from self.evaluate_entropy_string(
-                        chunk, line, string, HEX_CHARS, self.hex_entropy_limit
+                        chunk, line, string, self.hex_entropy_limit
                     )
 
     def evaluate_entropy_string(
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import abc
+from collections import Counter
 from functools import lru_cache
 import hashlib
 import logging
@@ -29,8 +30,8 @@ import pygit2
 from tartufo import config, types, util
 from tartufo.types import BranchNotFoundException, Rule, TartufoException
 
-BASE64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
-HEX_CHARS = "1234567890abcdefABCDEF"
+BASE64_REGEX = re.compile(r"[A-Z0-9+/]+={,2}", re.IGNORECASE)
+HEX_REGEX = re.compile(r"[0-9A-F]+", re.IGNORECASE)
 
 
 class Issue:
@@ -381,26 +382,26 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
         )
 
     @lru_cache(maxsize=None)
-    def calculate_entropy(self, data: str, char_set: str) -> float:
+    def calculate_entropy(self, data: str) -> float:
         """Calculate the Shannon entropy for a piece of data.
 
         This essentially calculates the overall probability for each character
-        in `data` to be to be present, based on the characters in `char_set`.
-        By doing this, we can tell how random a string appears to be.
+        in `data` to be to be present. By doing this, we can tell how random a
+        string appears to be.
 
         Borrowed from http://blog.dkbza.org/2007/05/scanning-data-for-entropy-anomalies.html
 
         :param data: The data to be scanned for its entropy
-        :param char_set: The character set used as a basis for the calculation
-        :return: The amount of entropy detected in the data.
+        :return: The amount of entropy detected in the data
         """
         if not data:
             return 0.0
+        frequency = Counter(data)
         entropy = 0.0
-        for char in char_set:
-            prob_x = float(data.count(char)) / len(data)
-            if prob_x > 0:
-                entropy += -prob_x * math.log2(prob_x)
+        float_size = float(len(data))
+        for count in frequency.values():
+            probability = float(count) / float_size
+            entropy += -probability * math.log2(probability)
         return entropy
 
     def scan(self) -> Generator[Issue, None, None]:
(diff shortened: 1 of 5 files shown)


tartufo/scanner.py
```diff
@@ -433,39 +434,33 @@ def scan_entropy(
...
         string: str,
-        chars: str,
         min_entropy_score: float,
     ) -> Generator[Issue, None, None]:
         """
         Check entropy string using entropy characters and score.
 
         :param chunk: The chunk of data to check
-        :param issues: Issue list to append any strings flagged
+        :param line: Source line containing string of interest
```
tarkatronic, line 490: Oh dang. Looks like this was totally out of date already, huh?

tartufo/scanner.py
```diff
@@ -345,26 +346,26 @@ def entropy_string_is_excluded(self, string: str, line: str, path: str) -> bool:
...
+        frequency = Counter(data)
         entropy = 0.0
-        for char in char_set:
-            prob_x = float(data.count(char)) / len(data)
-            if prob_x > 0:
-                entropy += -prob_x * math.log2(prob_x)
+        float_size = float(len(data))
+        for count in frequency.values():
+            probability = float(count) / float_size
+            entropy += -probability * math.log2(probability)
```
tarkatronic, line 404: So... I'm not saying this is necessarily good or bad, but this is a fundamental change to the way that we are calculating entropy. Instead of calculating how random a string is within a particular character set (base64 or hex), we are calculating it based on how random its characters look compared to themselves.

This idea, or at least very similar ideas, have been previously discussed, but never fully implemented:
* https://github.com/trufflesecurity/truffleHog/issues/168
* https://github.com/Yelp/detect-secrets/issues/234

I think this warrants FAR more investigation and discussion before we dive straight into it with both feet.

One idea that pops to mind is that perhaps we could introduce a new flag along the lines of `--experimental-entropy-detection` (or something less verbose) to let users test it out before we decide this is the right option.

tarkatronic (changes requested):
As noted in the review comments, I think we need a whole lot more investigation around this before steaming ahead with it.

Although the regex-based word splitting, if that can be pulled out into a discrete enhancement, that would clearly be faster!

tartufo/scanner.py
rbailey-godaddy: Actually, if you look at this closely, the math is bit-for-bit identical. I didn't put in a comment because it's weird enough without referencing code that doesn't even exist any longer, but here's the scoop:

The old algorithm:
1. Loops through each character in `char_set` (i.e. the encoding alphabet)
2. Counts the number of times that character appears in `data` (the target string)
3. Computes `prob_x` -- which makes me think "probability" although it isn't -- as the fraction of the input represented by this character (number of occurrences of character divided by length of target string)
4. Adds this to the running entropy tally using the gonzo equation from the literature

So the first observation is that if `char` doesn't appear in `data`, then `data.count(char)` is 0, which means `prob_x` is 0, which means `char` doesn't contribute to `entropy`.

The second observation is that since only characters in `data` contribute to `entropy`, it doesn't matter what the encoding alphabet is because only the characters actually in `data` count. Notice the only reference to `char_set` is in the `for` statement.

The third observation is that rather than blindly iterating the entire encoding alphabet and scanning the input for each character to count how many times it appears, it would be much better to scan the input once, tallying characters as we go, and then look only at the characters we know are present.

Thus, the new algorithm:
1. Uses `Counter` to quickly (or at least in fewer lines) count up which characters are present and how many times each was used
2. Computes the length of the target string as a float so we don't have to do a type promotion inside every loop pass
3. For each character that is present in `data`, compute the "probability" [sic - really frequency as a fraction of string size] just like before. Note it doesn't even matter what exactly the character is, which is why we iterate on `.values()` -- also note the original algorithm never used `char` (the loop index) except in `data.count()`.
4. Calculates the cumulative `entropy`. Note there is no test on `probability > 0` here because we know it's greater than 0 because we know `count` is greater than 0 because we know it's in `frequency` and we know `Counter` objects only have entries for things that have been counted (at least once).

Looked at differently, both iterate through all of the characters found in `data`...

OLD:
```
for char in [char in char_set if data.count(char) > 0]:
    prob_x = fraction-with-data.count(char)-in-numerator
    add prob_x to entropy
```

NEW:
```
for count in Counter(data).values():  # for char except we don't use char because we only care about count
    prob_x = fraction-with-data.count(char)-in-numerator aka count/size
    add prob_x to entropy
```

rbailey-godaddy: I was looking for this analogy the entire time I composed the above note, but it didn't come until after I posted it. Regarding the concern that we aren't looking at the representation alphabet any more (not that we were before) and the potential impact on the computation, this basically boils down to "what is the difference between a Big Mac without anchovies and a Big Mac without capers?" The answer, of course, is nothing.

The only reason we care about the alphabet at all -- and not here -- is to know how many bits of data each character represents, so that it is possible to scale the computed entropy value to get an apples-to-apples comparison of the underlying binary bitstream if we had bothered to convert it from base64 or hex representation into raw bits.

To illustrate: How random is the string "11111111"?

If this is a binary number, each digit has a 1-in-2 chance of being "1". If this is a hexadecimal representation, each digit has a 1-in-16 chance of being "1". Either way, it doesn't change the fact that all 8 of the digits are "1".

rbailey-godaddy: Regarding the independent investigations above, [trufflesecurity/truffleHog#168](https://github.com/trufflesecurity/truffleHog/issues/168) basically does the same derivation, but a bug in the open-coded Counter loop b0rks the results. I posted there with the code here and identified the bug.

Additionally, running the original truffleHog entropy, the proposed (there) single entropy, the old tartufo entropy and the proposed (here) new entropy on the sample strings shows that both originals and the new code proposed here yield exactly identical entropy scores, while the buggy code is, um, buggy.

[Yelp/detect-secrets](https://github.com/Yelp/detect-secrets/issues/234) addresses itself to a different problem, namely the question of what the default entropy detection threshold (aka "sensitivity") should be. This is indeed an interesting question, but is inapropos here to the extent that we maintain exactly the same default threshold as before (and compare the exact same entropy scores to it).

tarkatronic (approved):

PR #280, opened 2021-11-11
Version bump for v3.0.0-alpha.1

tarkatronic:
This will be the first alpha release of tartufo v3, to give users a taste of what is to come, and to let them help shake out any bugs.

Documentation on new features is sparse still, but I think this could prove a welcome glimpse.

Plus it gives users a chance to make sure they are pinned to a "safe" v2.x version before we get TOO far down this rabbit hole.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-vx.x.x - TBD
-------------
+v3.0.0-alpha.1 - 11 November 2021
+---------------------------------
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


sushantmimani (approved):

rbailey-godaddy (commented):
One initial sanity-check question first...

I'm not so familiar with the python (poetry/pip) versioning ecosystem, but in the rpm world this version would be problematic because `3.0.0-alpha.1` would come lexically AFTER `3.0.0` and we'd have to get to some number higher than 3.0.0 before ordering becomes sane again.

Will this version work the way we intend?

rbailey-godaddy (approved):

tarkatronic:
> One initial sanity-check question first...
> 
> I'm not so familiar with the python (poetry/pip) versioning ecosystem, but in the rpm world this version would be problematic because `3.0.0-alpha.1` would come lexically AFTER `3.0.0` and we'd have to get to some number higher than 3.0.0 before ordering becomes sane again.
> 
> Will this version work the way we intend?

For posterity sake, since this got answered in a private conversation, I just wanted to respond that yes! This absolutely works the way we would hope!

Poetry internally uses a `-alpha.1` indicator, which then gets translated when the distributable packages are built, to make the final "released" version equal to `3.0.0a1`. And this version number, as you can see here on PyPI, is understood to be a pre-release: https://pypi.org/project/tartufo/#history

For anybody interested in more information on the topic...

* Poetry follows the Semantic Versioning scheme laid out at semver.org, and then translates those to...
* The PEP on the Python versioning scheme: https://www.python.org/dev/peps/pep-0440/

PR #282, opened 2021-11-11
Remove old-style config for exclude-entropy-patterns

### LINKED ISSUE #264: Remove old-style config for exclude-entropy-patterns

As of v2.9.0 we have two ways of configuring `exclude-entropy-patterns` -- as a single string, and as a table of tables.

We should have only one definitive way of configuring this, and we are moving toward the "table of tables" syntax, so we should remove the ability to configure in the old manner.

sushantmimani:
Fix tests

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [x] Other

## Backward Compatibility

* [x] No (breaking changes)
Old style `exluded-entropy-patterns` will no longer be supported and will raise an error

## Issue Linking
Closes #264 

## What's new?
Old style `exluded-entropy-patterns` will no longer be supported and will raise an error. These must now follow the TOML `array of tables` format and can only be specified in the config file

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -280,9 +280,9 @@ digests. To avoid these false positives, enable ``exclude-entropy-patterns``. Ex
 apply to any strings flagged by entropy checks. This option is not available on the command line,
 and must be specified in your config file.
 
-For example, if ``docs/README.md`` contains a git SHA and `.github/workflows/*.yml` containes pinned git SHAs
+For example, if ``docs/README.md`` contains a git SHA and `.github/workflows/*.yml` contains pinned git SHAs
 this would be flagged by entropy.
-To exclude this, add an entry to ``exclude-entropy-patterns`` in the config file.
+To exclude these, add the following entries to ``exclude-entropy-patterns`` in the config file.
 
 .. code-block:: toml
 
@@ -291,6 +291,14 @@ To exclude this, add an entry to ``exclude-entropy-patterns`` in the config file
         {path-pattern = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
         {path-pattern = '\.github/workflows/.*\.yml', pattern = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
     ]
+.. note::
+    ``match-type`` is used to select the ``search`` or ``match`` regex operation. ``search`` looks for the regex
+    anywhere in the selected scope, while ``match`` requires the regex to match at the beginning of the selected scope.
+    Defaults to ``search``
+
+    ``scope`` is used to specify if you want to perform the regex operation (search or match) by ``word`` or ``line``.
+    ``word`` means exactly the high-entropy string of characters, while ``line`` searches the entire input line
+    containing the high-entropy string. Defaults to ``line``
 
 Thanks to the magic of TOML, you could also split these out into their own tables
 in the config if you wanted. So the following would be 100% equivalent to what
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -280,7 +280,8 @@ digests. To avoid these false positives, enable ``exclude-entropy-patterns``. Ex
 apply to any strings flagged by entropy checks. This option is not available on the command line,
 and must be specified in your config file.
 
-For example, if ``docs/README.md`` contains a git SHA, this would be flagged by entropy.
+For example, if ``docs/README.md`` contains a git SHA and `.github/workflows/*.yml` containes pinned git SHAs
+this would be flagged by entropy.
 To exclude this, add an entry to ``exclude-entropy-patterns`` in the config file.
 
 .. code-block:: toml
@@ -288,6 +289,7 @@ To exclude this, add an entry to ``exclude-entropy-patterns`` in the config file
     [tool.tartufo]
     exclude-entropy-patterns = [
         {path-pattern = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
+        {path-pattern = '\.github/workflows/.*\.yml', pattern = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
     ]
 
 Thanks to the magic of TOML, you could also split these out into their own tables
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -280,7 +280,7 @@ digests. To avoid these false positives, enable ``exclude-entropy-patterns``. Ex
 apply to any strings flagged by entropy checks. This option is not available on the command line,
 and must be specified in your config file.
 
-For example, if ``docs/README.md`` contains a git SHA and `.github/workflows/*.yml` contains pinned git SHAs
+For example, if ``docs/README.md`` contains a git SHA and ``.github/workflows/*.yml`` contains pinned git SHAs
 this would be flagged by entropy.
 To exclude these, add the following entries to ``exclude-entropy-patterns`` in the config file.
 
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -277,41 +277,17 @@ Entropy Limiting
 
 Entropy scans can produce a high number of false positives such as git SHAs or md5
 digests. To avoid these false positives, enable ``exclude-entropy-patterns``. Exclusions
-apply to any strings flagged by entropy checks.
+apply to any strings flagged by entropy checks. This option is not available on the command line,
+and must be specified in your config file.
 
 For example, if ``docs/README.md`` contains a git SHA, this would be flagged by entropy.
-To exclude this, add ``docs/.*\.md$::^[a-zA-Z0-9]{40}$`` to ``exclude-entropy-patterns``.
-
-.. code-block:: sh
-
-    > tartufo ... --exclude-entropy-patterns "docs/.*\.md$::^[a-zA-Z0-9]{40}$"
-
-.. code-block:: toml
-
-    [tool.tartufo]
-    exclude-entropy-patterns = [
-      # format: "{file regex}::{entropy pattern}"
-      "docs/.*\.md$::^[a-zA-Z0-9]{40}$", # exclude all git SHAs in the docs directory
-    ]
-
-.. warning::
-    .. versionchanged:: 2.9.0
-    As of version 2.9.0, the above specification style has been deprecated, and
-    will be removed in version 3.0. The new style uses a TOML `array of tables`_
-    as shown below.
-
-    Note that this new syntax is not available on the command line, and must be
-    specified in your config file.
-
-Here is an example of how you might exclude SHA hashes in your docs, as well as
-hashes for GitHub Actions in your workflows:
+To exclude this, add an entry to ``exclude-entropy-patterns`` in the config file.
 
 .. code-block:: toml
 
     [tool.tartufo]
     exclude-entropy-patterns = [
         {path-pattern = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
-        {path-pattern = '\.github/workflows/.*\.yml', pattern = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
     ]
 
 Thanks to the magic of TOML, you could also split these out into their own tables
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -120,10 +120,11 @@ class TartufoCLI(click.MultiCommand):
     "-xe",
     "--exclude-entropy-patterns",
     multiple=True,
-    help="""Specify a regular expression which matches entropy strings to
-    exclude from the scan. This option can be specified multiple times to
-    exclude multiple patterns. If not provided (default), no entropy strings
-    will be excluded ({path regex}::{pattern regex}).""",
+    hidden=True,
+    help="""Specify a regular expression which matches entropy strings to exclude from the scan. This option can be
+    specified multiple times to exclude multiple patterns. If not provided (default), no entropy strings will be
+    excluded. ({"path-pattern": {path regex}, "pattern": {pattern regex}, "match-type": "match"|"search",
+    "scope": "word"|"line"}).""",
 )
 @click.option(
     "-e",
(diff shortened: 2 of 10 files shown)


rbailey-godaddy (approved):
LGTM

docs/source/features.rst
```diff
@@ -277,42 +277,24 @@ Entropy Limiting
...
 
     [tool.tartufo]
     exclude-entropy-patterns = [
-      # format: "{file regex}::{entropy pattern}"
-      "docs/.*\.md$::^[a-zA-Z0-9]{40}$", # exclude all git SHAs in the docs directory
+        {path-pattern = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'
+         match-type = 'search' , scope = 'line' },
     ]
+.. note::
+    ``match-type`` is used to select the ``search`` or ``match`` regex operation. Defaults to ``search``
```
rbailey-godaddy:
```suggestion
    ``match-type`` is used to select the ``search`` or ``match`` regex operation. ``search`` looks for the regex anywhere in
    the selected scope, while ``match`` requires the regex to match at the beginning of the selected scope. Defaults to ``search``
```
This could be worded better than my suggestion. I'm unsure that all users (even if they know python) will understand the distinction without some additional help.

docs/source/features.rst
```diff
@@ -277,42 +277,24 @@ Entropy Limiting
...
-
-.. code-block:: toml
-
-    [tool.tartufo]
-    exclude-entropy-patterns = [
-        {path-pattern = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
-        {path-pattern = '\.github/workflows/.*\.yml', pattern = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
-    ]
+    ``scope`` is used to specify if you want to perform the regex operation (search or match) by ``word`` or ``line``.
+    Defaults to ``line``
```
rbailey-godaddy:
```suggestion
    ``scope`` is used to specify if you want to perform the regex operation (search or match) by ``word`` or ``line``.
    ``word`` means exactly the high-entropy string of characters, while ``line`` searches the entire input line containing the high-entropy string.
    Defaults to ``line``

```
Basically the same commentary as above

tartufo/cli.py
```diff
@@ -120,10 +120,10 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
     "-xe",
     "--exclude-entropy-patterns",
     multiple=True,
-    help="""Specify a regular expression which matches entropy strings to
-    exclude from the scan. This option can be specified multiple times to
-    exclude multiple patterns. If not provided (default), no entropy strings
-    will be excluded ({path regex}::{pattern regex}).""",
+    help="""Specify a regular expression which matches entropy strings to exclude from the scan. This option can be
+    specified multiple times to exclude multiple patterns. If not provided (default), no entropy strings will be
+    excluded. ({"path-pattern": {path regex}, "pattern": {pattern regex}, "match-type": "match"|"search",
```
rbailey-godaddy, line 126:
```suggestion
    specified multiple times to exclude multiple patterns. If not provided (default), all high-entropy strings will be
    reported. ({"path-pattern": {path regex}, "pattern": {pattern regex}, "match-type": "match"|"search",
```

rbailey-godaddy (approved):
It still looks good. ;-) I made some nit-picking editorial comments on the documentation, which you can take or leave.

docs/source/features.rst
```diff
@@ -277,41 +277,17 @@ Entropy Limiting
...
-Here is an example of how you might exclude SHA hashes in your docs, as well as
-hashes for GitHub Actions in your workflows:
+To exclude this, add an entry to ``exclude-entropy-patterns`` in the config file.
 
 .. code-block:: toml
 
     [tool.tartufo]
     exclude-entropy-patterns = [
         {path-pattern = 'docs/.*\.md$', pattern = '^[a-zA-Z0-9]$', reason = 'exclude all git SHAs in the docs'},
-        {path-pattern = '\.github/workflows/.*\.yml', pattern = 'uses: .*@[a-zA-Z0-9]{40}', reason = 'GitHub Actions'}
```
tarkatronic, line 314: Why remove this second example? I think it's better to show with multiple entries here. :)

tarkatronic: Maybe just update the paragraph above to say something like, "Imagine, for example, you want to exclude all git SHAs in a docs directory, as well as references to external GitHub Actions versions in your workflows." -- more wordsmithing should certainly be applied, but something like that 😄

tartufo/config.py
```diff
@@ -217,15 +216,19 @@ def load_rules_from_file(rules_file: TextIO) -> Dict[str, Rule]:
             rule = Rule(
                 name=rule_name,
                 pattern=re.compile(rule_definition["pattern"]),
-                path_pattern=re.compile(path_pattern) if path_pattern else None,
+                path_pattern=re.compile(path_pattern)
+                if path_pattern
+                else re.compile(""),
```
tarkatronic: Super minor nit: Can we move this up to a constant, `EMPTY_PATTERN = re.compile("")`? That way we don't have to potentially recompile an empty string regex many times. Even if it is a super low-cost operation, might as well minimize it.

tartufo/types.py
```diff
@@ -74,18 +74,29 @@ class Chunk:
...
 
 
+class MatchType(enum.Enum):
+    Match = "match"  # pylint: disable=invalid-name
+    Search = "search"  # pylint: disable=invalid-name
+
+
+class Scope(enum.Enum):
+    Word = "word"  # pylint: disable=invalid-name
+    Line = "line"  # pylint: disable=invalid-name
```
tarkatronic: If we're creating these enums (which I certainly support) we should make sure that they are used throughout `config.py` and `scanner.py` where appropriate!

tartufo/util.py
```diff
@@ -50,13 +48,6 @@ def del_rw(_func: Callable, name: str, _exc: Exception) -> None:
     os.remove(name)
 
 
-def convert_regexes_to_rules(regexes: Dict[str, Pattern]) -> Dict[str, Rule]:
-    return {
-        name: Rule(name=name, pattern=pattern, path_pattern=None, re_match_type="match")
-        for name, pattern in regexes.items()
-    }
```
tarkatronic, line 57: Are you certain we don't still need this? Isn't this how the `default_regexes.json` gets loaded? https://github.com/godaddy/tartufo/blob/main/tartufo/data/default_regexes.json

tarkatronic (changes requested):
Nothing too serious. But I want to be certain we retain the old regex loading functionality for the `default_regexts.json` loading. And either utilize the new `MatchType` and `Scope` enums or remove them since they're not currently used at all. 😄

tartufo/util.py
sushantmimani: Yes, `convert_regexes_to_rules` has 0 references. `default_regexes.json` gets loaded in `configure_regexes`

docs/source/features.rst
```diff
@@ -277,34 +277,12 @@ Entropy Limiting
...
-    As of version 2.9.0, the above specification style has been deprecated, and
-    will be removed in version 3.0. The new style uses a TOML `array of tables`_
-    as shown below.
-
-    Note that this new syntax is not available on the command line, and must be
-    specified in your config file.
-
-Here is an example of how you might exclude SHA hashes in your docs, as well as
-hashes for GitHub Actions in your workflows:
+For example, if ``docs/README.md`` contains a git SHA and `.github/workflows/*.yml` containes pinned git SHAs
```
tarkatronic:
```suggestion
For example, if ``docs/README.md`` contains a git SHA and `.github/workflows/*.yml` contains pinned git SHAs
```

docs/source/features.rst
```diff
@@ -277,34 +277,12 @@ Entropy Limiting
...
+this would be flagged by entropy.
+To exclude this, add an entry to ``exclude-entropy-patterns`` in the config file.
```
tarkatronic:
```suggestion
To exclude these, add an entry to ``exclude-entropy-patterns`` in the config file.
```

tartufo/config.py
```diff
@@ -279,25 +262,12 @@ def compile_rules(patterns: Iterable[Union[str, Dict[str, str]]]) -> List[Rule]:
                 Rule(
                     name=pattern.get("reason", None),  # type: ignore[union-attr]
                     pattern=re.compile(pattern["pattern"]),  # type: ignore[index]
-                    path_pattern=re.compile(pattern.get("path-pattern", ".*")),  # type: ignore[union-attr]
-                    re_match_type="search",
+                    path_pattern=re.compile(pattern.get("path-pattern", "")),  # type: ignore[union-attr]
+                    re_match_type=pattern.get("match-type", MatchType.Search),  # type: ignore[union-attr]
+                    re_match_scope=pattern.get("scope", Scope.Line),  # type: ignore[union-attr]
```
tarkatronic: We should ensure that these are valid values. You could do something like this...

```python
try:
    match_type = MatchType(pattern.get("match-type", MatchType.Search.value))
except ValueError as exc:
    raise ConfigException(f"Invalid value for match-type: {pattern.get('match-type')}") from exc
```
Of course, that breaks the list comprehension... but we do want to be sure that they don't supply something weird like `match-type = "ford-prefect"`, or just typo like `match-type = "saerch"`.

tartufo/scanner.py
```diff
@@ -351,18 +357,23 @@ def rule_matches(rule: Rule, string: str, line: str, path: str) -> bool:
...
         :param string: string to match against rule pattern
+        :param line: Source line containing string of interest
         :param path: path to match against rule path_pattern
         :return: True if string and path matched, False otherwise.
         """
         match = False
-        if rule.re_match_type == "match":
+        if rule.re_match_scope == Scope.Word:
+            scope = string
+        else:
```
tarkatronic: We should probably explicitly match `Scope.Line` here and error out for invalid values.

tartufo/scanner.py
```diff
@@ -351,18 +357,23 @@ def rule_matches(rule: Rule, string: str, line: str, path: str) -> bool:
...
+                match = rule.pattern.match(scope) is not None
             if rule.path_pattern:
                 match = match and rule.path_pattern.match(path) is not None
-        elif rule.re_match_type == "search":
+        elif rule.re_match_type == MatchType.Search:
             if rule.pattern:
-                match = rule.pattern.search(line) is not None
+                match = rule.pattern.search(scope) is not None
             if rule.path_pattern:
                 match = match and rule.path_pattern.search(path) is not None
```
tarkatronic, line 380: Similar to above, we should probably error out for invalid values now that we're formalizing the ability to configure this functionality.

docs/source/features.rst
```diff
@@ -277,34 +277,12 @@ Entropy Limiting
...
 
 .. code-block:: toml
```
tarkatronic, line 288: We should add documentation of all the available keys, which ones are optional, and what they default to.

tarkatronic (changes requested):

tarkatronic:
Another thought here: We've now documented that this option is only available in the config file, and not from the CLI. But it's still being added as a `@cli.option`. So when the user runs `tartufo --help` they will see it as a potential option on the command line. We should see if we can remove it from there, and still use it as a config-file option.

tartufo/config.py
sushantmimani: If I add checks for but `MatchType` and `Scope` here, I can avoid checking for values for `Scope` in `scanner.py`, right?

docs/source/features.rst
sushantmimani: I had added these but they were reverted back in one of the rebases. Added them back now.

docs/source/features.rst
```diff
@@ -277,34 +277,12 @@ Entropy Limiting
...
-    As of version 2.9.0, the above specification style has been deprecated, and
-    will be removed in version 3.0. The new style uses a TOML `array of tables`_
-    as shown below.
-
-    Note that this new syntax is not available on the command line, and must be
-    specified in your config file.
-
-Here is an example of how you might exclude SHA hashes in your docs, as well as
-hashes for GitHub Actions in your workflows:
+For example, if ``docs/README.md`` contains a git SHA and `.github/workflows/*.yml` contains pinned git SHAs
```
tarkatronic:
```suggestion
For example, if ``docs/README.md`` contains a git SHA and ``.github/workflows/*.yml`` contains pinned git SHAs
```
Making the highlighting consistent 😄

tartufo/config.py
```diff
@@ -246,58 +250,38 @@ def compile_path_rules(patterns: Iterable[str]) -> List[Pattern]:
...
-        )
-        stripped = (p.strip() for p in patterns)  # type: ignore[union-attr]
-        rules = [
-            compile_rule(pattern)
-            for pattern in stripped
-            if pattern and not pattern.startswith("#")
-        ]
-        return rules
+            )
+        except (KeyError, AttributeError) as exc:
```
tarkatronic: Can `AttributeError` still occur here? I'm looking at it and I can't spot where it might happen. Am I missing something?

tartufo/scanner.py
```diff
@@ -351,18 +357,23 @@ def rule_matches(rule: Rule, string: str, line: str, path: str) -> bool:
...
+        :param line: Source line containing string of interest
         :param path: path to match against rule path_pattern
         :return: True if string and path matched, False otherwise.
         """
         match = False
-        if rule.re_match_type == "match":
+        if rule.re_match_scope == Scope.Word:
+            scope = string
+        elif rule.re_match_scope == Scope.Line:
+            scope = line
```
tarkatronic, line 368: We should have an `else` block to explicitly error out here. Otherwise, if a user utilizing this library passes an invalid value (let's say a `Rule` with a match_scope of `None`), we will get a `NameError` when further down when you try to reference `scope`.

tarkatronic (changes requested):
Just a couple of small updates left, viewing this from the perspective of being a library/API first, with a CLI exposing it second. Great work @sushantmimani!

tartufo/config.py
sushantmimani: It cannot. This was leftover from previous changes. Removed now.

tarkatronic (approved):
🎉 🚢 :shipit:

PR #283, opened 2021-11-11
Adding --recurse / --no-recurse option

### LINKED ISSUE #268: Add a `--recurse` / `--no-recurse` flag to `scan-folder`

Currently, the `scan-folder` always recurses down into all sub-folders of the target folder for its scan. It would be nice to be able to control that behavior and do only a top-level scan if preferred. This could be especially beneficial on particularly large folders.

A new boolean flag should be added: `--recurse` / `--no-recurse`, defaulted to `--recurse`, to maintain current behavior.

mayuriesha:

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
Closes #268 

## What's new?
- Introducing new flag --recurse/--no-recurse to scan entire directory or just the root directory. It is defaulted to recurse to maintain the current behavior of scan-folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ Features:
 * [#254](https://github.com/godaddy/tartufo/pull/260) - Changes the default value of
   `--regex/--no-regex` to True.
 * [#268](https://github.com/godaddy/tartufo/issues/268) - Adds a new
-  `--recurse / --no-recurse` flag which allows users scan entire directory or just the root directory
+  `--recurse / --no-recurse` flag which allows users to recursively scan the entire directory or just
+  the root directory
 
 Misc:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ Features:
   `--scan-filenames/--no-scan-filenames` flag which allows users to enable or disable file name scanning.
 * [#254](https://github.com/godaddy/tartufo/pull/260) - Changes the default value of
   `--regex/--no-regex` to True.
+* [#268](https://github.com/godaddy/tartufo/issues/268) - Adds a new
+  `--recurse / --no-recurse` flag which allows users scan entire directory or just the root directory
 
 Misc:
 
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ Commands:
   pre-commit        Scan staged changes in a pre-commit hook.
   scan-local-repo   Scan a repository already cloned to your local system.
   scan-remote-repo  Automatically clone and scan a remote git repository.
+  scan-folder       Scan a folder.
 
 ```
 
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ Commands:
   pre-commit        Scan staged changes in a pre-commit hook.
   scan-local-repo   Scan a repository already cloned to your local system.
   scan-remote-repo  Automatically clone and scan a remote git repository.
-  scan-folder       Scan a folder.
+  scan-folder       Scan a folder on your local system.
 
 ```
 
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -761,9 +761,7 @@ class FolderScanner(ScannerBase):
     target: str
 
     def __init__(
-        self,
-        global_options: types.GlobalOptions,
-        target: str,
+        self, global_options: types.GlobalOptions, target: str, recurse: bool
     ) -> None:
         """Used for scanning a folder.
 
@@ -771,6 +769,7 @@ class FolderScanner(ScannerBase):
         :param target: The local filesystem path to scan
         """
         self.target = target
+        self.recurse = recurse
         super().__init__(global_options)
 
     @property
@@ -785,7 +784,8 @@ class FolderScanner(ScannerBase):
 
     def _iter_folder(self) -> Generator[Tuple[str, str], None, None]:
         folder_path = pathlib.Path(self.target)
-        for file_path in folder_path.rglob("**/*"):
+        files = folder_path.rglob("**/*") if self.recurse else folder_path.glob("*")
+        for file_path in files:
             relative_path = file_path.relative_to(folder_path)
             if file_path.is_file() and self.should_scan(str(relative_path)):
                 try:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -759,6 +759,7 @@ class FolderScanner(ScannerBase):
     """Used to scan a folder."""
 
     target: str
+    recurse: bool
 
     def __init__(
         self, global_options: types.GlobalOptions, target: str, recurse: bool
--- a/tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt
+++ b/tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt
@@ -1,0 +1,3 @@
+not matched
+ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=
+another not matched
--- a/tests/test_folder_scanner.py
+++ b/tests/test_folder_scanner.py
@@ -64,8 +64,7 @@ class FolderScannerTestCase(unittest.TestCase):
         actual_issues = []
         for issue in issues:
             actual_issues.append(issue.matched_string)
-        self.assertIn("KQ0I97OBuPlGB9yPRxoSxnX52zE=", actual_issues)
-        self.assertIn("KQ0I97OBuPlGB9yPRxoSxnX52zE=", actual_issues)
+        self.assertEqual(2, actual_issues.count("KQ0I97OBuPlGB9yPRxoSxnX52zE="))
         self.assertIn(
             "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=",
             actual_issues,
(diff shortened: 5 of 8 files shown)


rbailey-godaddy (approved):
If you don't mind a little more tweaking...

It would be more consistent with existing behavior to look at the value of `global_options.recurse` instead of extracting it in the caller and then passing `recurse` as a separate parameter. Additionally, you can then just check `self.global_options.recurse` within the class instead of needing to add `self.recurse`.

Also, the more I look at this, the more I can't see where the `recurse` parameter to `scan_folder.main()` is coming from. The crazy thing is that I play-tested it and it obviously works, so I guess I'm good with things as-is if nobody else has complaints -- I just can't understand what is happening!

mayuriesha:
I believe `self.global_options` are for the flags that are defined in cli.py which are` tartufo command line options`. Where as the `recurse` is an option that goes with sub_command `scan-folder`. So I am not sure if we can use `self.global_options.recurse`.

About the parameter `recurse` for `scan_folder->main`, as far as I understand the `click.command` is handling the arguments/options that are defined and passes it to the method definition below.
https://github.com/pallets/click/blob/5d759e0ab383136f9869cdb5e14f00c78b38a9c9/docs/options.rst

README.md
```diff
@@ -159,6 +159,7 @@ Commands:
   pre-commit        Scan staged changes in a pre-commit hook.
   scan-local-repo   Scan a repository already cloned to your local system.
   scan-remote-repo  Automatically clone and scan a remote git repository.
+  scan-folder       Scan a folder.
```
sushantmimani: nit
```suggestion
  scan-folder       Scan a folder on your local system.
```

tartufo/scanner.py
```diff
@@ -749,16 +749,15 @@ class FolderScanner(ScannerBase):
...
-        target: str,
+        self, global_options: types.GlobalOptions, target: str, recurse: bool
     ) -> None:
         """Used for scanning a folder.
 
         :param global_options: The options provided to the top-level tartufo command
         :param target: The local filesystem path to scan
         """
         self.target = target
+        self.recurse = recurse
```
sushantmimani, line 818: Declare recurse after L749 for consistency.

tests/test_folder_scanner.py
```diff
@@ -16,30 +16,58 @@ def setUp(self) -> None:
...
         self.global_options.exclude_path_patterns = [r"donotscan\.txt"]
 
-        test_scanner = scanner.FolderScanner(self.global_options, folder_path)
+        test_scanner = scanner.FolderScanner(self.global_options, folder_path, recurse)
         issues = list(test_scanner.scan())
 
-        self.assertEqual(1, len(issues))
+        self.assertEqual(2, len(issues))
         self.assertEqual("KQ0I97OBuPlGB9yPRxoSxnX52zE=", issues[0].matched_string)
         self.assertEqual(IssueType.Entropy, issues[0].issue_type)
```
sushantmimani: Need a similar assertions for the new issue.

tartufo/scanner.py
```diff
@@ -749,16 +749,15 @@ class FolderScanner(ScannerBase):
     target: str
 
     def __init__(
-        self,
-        global_options: types.GlobalOptions,
-        target: str,
+        self, global_options: types.GlobalOptions, target: str, recurse: bool
```
sushantmimani, line 810: If we default this to true here, we can avoid changing a bunch of tests
```suggestion
        self, global_options: types.GlobalOptions, target: str, recurse: bool = True
```

But then again, we don't do this for options like `include-submodules` so I'll defer to other about whether we should do this or not.

tests/test_folder_scanner.py
```diff
@@ -16,30 +16,58 @@ def setUp(self) -> None:
...
+        recurse = False
+        self.global_options.entropy = True
+        self.global_options.exclude_signatures = []
+        self.global_options.b64_entropy_score = 4.5
+        self.global_options.hex_entropy_score = 3
+
+        test_scanner = scanner.FolderScanner(self.global_options, folder_path, recurse)
+        issues = list(test_scanner.scan())
+
+        self.assertEqual(2, len(issues))
```
sushantmimani, line 82: We should assert that the issues found are what we're expecting.

tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt
```diff
@@ -0,0 +1,3 @@
+not matched
+ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=
```
sushantmimani, line 2: This requires a new signature to be added to `pyproject.toml`. I'd recommend installing pre-commit, if you haven't already, so that issues such as these are caught before pushing to GH.

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ Features:
   `--scan-filenames/--no-scan-filenames` flag which allows users to enable or disable file name scanning.
 * [#254](https://github.com/godaddy/tartufo/pull/260) - Changes the default value of
   `--regex/--no-regex` to True.
+* [#268](https://github.com/godaddy/tartufo/issues/268) - Adds a new
+  `--recurse / --no-recurse` flag which allows users scan entire directory or just the root directory
```
sushantmimani:
```suggestion
  `--recurse / --no-recurse` flag which allows users to recursively scan the entire directory or just the root directory
```

sushantmimani (changes requested):

tartufo/scanner.py
mayuriesha: I will skip this change to maintain consistency for now unless we want to have this change.

sushantmimani: @jwilhelm-godaddy  @rbailey-godaddy What are your thoughts about this?

tarkatronic: I think it's fine to leave this as-is. tartufo has been built as a library first, with a CLI that just happens to use that library. With that in mind, the library tries to operate with as few assumptions as possible. So if we can avoid an assumption here and remain more explicit, I would say that's preferable.

sushantmimani (approved):
LGTM 🦅

tests/test_folder_scanner.py
```diff
@@ -16,30 +16,77 @@ def setUp(self) -> None:
...
-        test_scanner = scanner.FolderScanner(self.global_options, folder_path)
+        test_scanner = scanner.FolderScanner(self.global_options, folder_path, recurse)
         issues = list(test_scanner.scan())
 
-        self.assertEqual(1, len(issues))
-        self.assertEqual("KQ0I97OBuPlGB9yPRxoSxnX52zE=", issues[0].matched_string)
+        self.assertEqual(2, len(issues))
+        actual_issues = []
+        for issue in issues:
+            actual_issues.append(issue.matched_string)
```
tarkatronic: Not a blocker, but cases like this are ideal for list comprehensions 😄 
```suggestion
        actual_issues = [issue.matched_string for issue in issues]
```

tests/test_folder_scanner.py
```diff
@@ -16,30 +16,77 @@ def setUp(self) -> None:
...
 
-        test_scanner = scanner.FolderScanner(self.global_options, folder_path)
+        test_scanner = scanner.FolderScanner(self.global_options, folder_path, recurse)
 
         with patch("pathlib.Path.open", side_effect=OSError()):
             with self.assertRaises(click.FileError):
                 list(test_scanner.scan())
 
+    def test_scan_all_the_files_recursively(self):
+        folder_path = pathlib.Path(__file__).parent / "data/scan_folder"
```
tarkatronic: All path separators should be outside the string, so that we can be absolutely certain they are OS-independent.

```suggestion
        folder_path = pathlib.Path(__file__).parent / "data" / "scan_folder"
```

tests/test_folder_scanner.py
```diff
@@ -16,30 +16,77 @@ def setUp(self) -> None:
...
+        self.global_options.b64_entropy_score = 4.5
+        self.global_options.hex_entropy_score = 3
+
+        test_scanner = scanner.FolderScanner(self.global_options, folder_path, recurse)
+        issues = list(test_scanner.scan())
+
+        self.assertEqual(3, len(issues))
+        actual_issues = []
+        for issue in issues:
+            actual_issues.append(issue.matched_string)
```
tarkatronic: Same list comprehension trick can be applied here
```suggestion
        actual_issues = [issue.matched_string for issue in issues]
```

tests/test_folder_scanner.py
```diff
@@ -16,30 +16,77 @@ def setUp(self) -> None:
...
+        self.assertEqual(2, actual_issues.count("KQ0I97OBuPlGB9yPRxoSxnX52zE="))
+        self.assertIn(
+            "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=",
+            actual_issues,
+        )
+
+    def test_scan_only_root_level_files(self):
+        folder_path = pathlib.Path(__file__).parent / "data/scan_folder"
```
tarkatronic:
```suggestion
        folder_path = pathlib.Path(__file__).parent / "data" / "scan_folder"
```

tests/test_folder_scanner.py
```diff
@@ -16,30 +16,77 @@ def setUp(self) -> None:
...
+        self.global_options.b64_entropy_score = 4.5
+        self.global_options.hex_entropy_score = 3
+
+        test_scanner = scanner.FolderScanner(self.global_options, folder_path, recurse)
+        issues = list(test_scanner.scan())
+
+        self.assertEqual(2, len(issues))
+        actual_issues = []
+        for issue in issues:
+            actual_issues.append(issue.matched_string)
```
tarkatronic:
```suggestion
        actual_issues = [issue.matched_string for issue in issues]
```

tarkatronic (approved):
No hard blockers here; overall it looks great! Just a few suggestions to improve the tests with some Python-isms. 😄 

Thank you!

Issue #257, opened 2021-11-02
Change `--include-path-patterns` and `--exclude-path-patterns` to use a table of tables


tarkatronic:
## Is your feature request related to a problem? Please describe.

With the [new style entropy exclusions](https://tartufo.readthedocs.io/en/stable/features.html#entropy-limiting) using the table of tables syntax, and exclusion reasons [adopting the same](https://github.com/godaddy/tartufo/issues/202#issuecomment-897995597), we should work to unify the syntax where applicable.

## Describe the solution you'd like

The `--include-path-patterns` and `--exclude-path-patterns` configuration should become a tables of tables with the keys `path-pattern` and `reason`.

tarkatronic:
Fixed in #289

PR #286, opened 2021-11-20
Exclusion reasons

### LINKED ISSUE #202: Exclusion Reasons

It took me a while to generate an exclusion report that explains what is excluded and why. One thing I’d love to see in tartufo is the ability to specify the reason along-side the exclusion and have a report produced that we can easily provide to a security team. When someone adds an exclusion, it might not be front-and-center we need to track the reason until it’s time to create an exclusion report. We need to reverse engineer the exclusions to determine the reason.

A reason field along-side any exclusion.

`tartufo --json ...` along with manually generating the report.

Files

```text
tartufo.toml::contains tartufo signatures flagged by entropy
```

Config

```text
exclude-signatures=[
   "dfg99df8g0d98fg0s9d8fg09d8sfgd90sfg8s0df9::characters in google doc url"
]
exclude-entropy=[
  "file1.md::^[a-zA-Z0-9]{40}$::git sha in a markdown file"
}
```

mayuriesha:

* [ ] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
closes #202 

## What's new?
- Supports new format of exclusions in config file with the ability to specify the reason along with exclusion

### DIFF
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -355,7 +355,8 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
         if signatures:
             warnings.warn(
                 "--exclude-signatures will be deprecated. Make sure all the exclusions are moved to "
-                "exclude-findings section with new format.",
+                "exclude-findings section with new format. Example: "
+                "exclude-findings = [{signature='signature', reason='The reason of excluding the signature'}]",
                 DeprecationWarning,
             )
             configured_signatures.extend(signatures)
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -354,8 +354,8 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
         signatures = self.config_data.get("exclude_signatures", None)
         if signatures:
             warnings.warn(
-                "--exclude-signatures will be deprecated. Make sure all the exclusions are moved to "
-                "exclude-findings section with new format. Example: "
+                "--exclude-signatures has been deprecated and will be removed in a future version. "
+                "Make sure all the exclusions are moved to exclude-findings section with new format. Example: "
                 "exclude-findings = [{signature='signature', reason='The reason of excluding the signature'}]",
                 DeprecationWarning,
             )
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -150,7 +150,7 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
     global_options: types.GlobalOptions
     logger: logging.Logger
     _scan_lock: threading.Lock = threading.Lock()
-    _excluded_findings: tuple = ()
+    _excluded_findings: Tuple[str, ...] = ()
     _config_data: MutableMapping[str, Any] = {}
 
     def __init__(self, options: types.GlobalOptions) -> None:
@@ -344,9 +344,8 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
         return self._config_data
 
     @config_data.setter
-    def config_data(self, data) -> MutableMapping[str, Any]:
+    def config_data(self, data: MutableMapping[str, Any]) -> None:
         self._config_data = data
-        return self._config_data
 
     @cached_property
     def excluded_findings(self) -> tuple:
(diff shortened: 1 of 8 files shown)


tartufo/scanner.py
```diff
@@ -336,6 +339,36 @@ def should_scan(self, file_path: str):
...
+        self._config_data = data
+        return self._config_data
+
+    @cached_property
+    def excluded_findings(self) -> tuple:
+        configured_signatures = []
+        signatures = self.config_data.get("exclude_signatures", None)
+        if signatures:
+            warnings.warn(
+                "--exclude-signatures will be deprecated. Make sure all the exclusions are moved to "
```
```suggestion
                "--exclude-signatures has been deprecated and will be removed in a future version. Make sure all the exclusions are moved to "
```

tartufo/scanner.py
```diff
@@ -336,6 +339,36 @@ def should_scan(self, file_path: str):
...
         return True
 
+    @property
+    def config_data(self):
+        return self._config_data
+
+    @config_data.setter
+    def config_data(self, data) -> MutableMapping[str, Any]:
+        self._config_data = data
+        return self._config_data
```
tarkatronic: I can't quite figure out why this property is needed. Can you explain a bit?

tests/test_git_repo_scanner.py
```diff
@@ -93,21 +93,6 @@ def test_extra_exclusions_get_added(self, mock_load: mock.MagicMock):
...
-            {"exclude_signatures": ["foo", "bar"]},
-        )
-        test_scanner = scanner.GitRepoScanner(
-            self.global_options, self.git_options, str(self.data_dir)
-        )
-        test_scanner.load_repo("../tartufo")
-        self.assertEqual(
-            sorted(test_scanner.global_options.exclude_signatures), ["bar", "foo"]
-        )
-
```
tarkatronic, line 110: Why has this test been removed? We still need to be certain that signatures from the target repository are loaded and respected.

tartufo/scanner.py
mayuriesha: We need this as a place holder for the config data, so that we can use it to compute the signature as part of excluded_findings which is accessed without passing inputs.

tests/test_git_repo_scanner.py
mayuriesha: We are no longer computing signatures as part of load_repo and instead it will be computed whenever excluded_findings is called.

tarkatronic: Are you saying that the signatures are no longer loaded from the remote repo? Because that is necessary functionality. This test is here to ensure that, when a repository is loaded, we find any signatures configured in that repository, and add them to list of signatures already configured. It doesn't matter so much _how_ that's accomplished, but it does matter that it does happen.

tartufo/scanner.py
```diff
@@ -149,6 +150,8 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
     global_options: types.GlobalOptions
     logger: logging.Logger
     _scan_lock: threading.Lock = threading.Lock()
+    _excluded_findings: tuple = ()
```
```suggestion
    _excluded_findings: Tuple[str, ...] = ()
```

tartufo/scanner.py
```suggestion
    def config_data(self, data: MutableMapping[str, Any]) -> None:
        self._config_data = data
```
Returning data from a setter in Python is generally not a valuable operation, because there's no way to access it. You would need something like...

```python
new_data = scanner.config_data = data
```
Which is just confusing. And probably won't even work the way you expect.

You can find a bit more of an explanation here, if you're interested: https://stackoverflow.com/a/16615910

tarkatronic (approved):
This is great @mayuriesha, thank you!

As we have continued talking about this feature, and #257, I can't help but wonder if I was wrong in thinking we should allow both old-style and new-style config to exist side-by-side. The more I think about it, the more I think I was wrong. We're already forcing so much config change and churn on our users, why force them to learn new names for the config entries as well, especially names that might not make as much sense?

I hate to suggest a change to this so late. But I wonder if this, and #257, are cases where we should operate similar to what we did with `exclude-entropy-patterns` in 2.x. Either you use all old-style config, or all new-style config. I am honestly open to going either direction, so I'd love some more thoughts on it! And of course we can always merge this PR as-is, to get it checked off, and update it in the future if we decide to go the other way!

Thank you again for the hard work @mayuriesha!

PR #287, opened 2021-11-22
Deprecate --rules in favor of rule-patterns

### LINKED ISSUE #256: Deprecate `--rules` in favor of `--rule-patterns`

At this point, all tartufo config has been moved into the one central configuration file, except for `--rules`. In order to keep everything consistent, we should fix this one last one.

* Use of the `--rules` option should present the user with a `DeprecationWarning`
* New regex patterns should instead be specified in the configuration file as a table of tables, similar to [new style entropy exclusion](https://tartufo.readthedocs.io/en/stable/features.html#entropy-limiting). It would probably look something like

    ```toml
    [tool.tartufo]
    rule-patterns = [
      {name = "Private key", pattern = "---- BEGIN EC PRIVATE KEY ----", path_pattern = "/some/path"}
    ]
    ```
    OR
    ```toml
    [[tool.tartufo.rule-patterns]]
    name = "Private key"
    pattern = "---- BEGIN EC PRIVATE KEY ----"
    path_pattern = "/some/path"
    ```
* Required keys would be "name" and "pattern"; "path_pattern" would be optional.

As far as the future deprecation path, I see two possibilities:

1. This gets added to `main` directly, then ported to the `v3.x` branch, with the `--rules` option removed entirely in v3.0
2. This gets merged into `v3.x`, and we decide upon some far future version (likely even v4.0) where the `--rules` option can be fully removed

The choice is up to the author of the PR!

tarkatronic:

* [X] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [X] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking

Fixes #256 

## What's new?
- A new `rule-patterns` configuration option has been added to move the contents of rules files directly into the tartufo config file.
- Using the old `--rules` option will now raise a `DeprecationWarning`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ Features:
 * [#268](https://github.com/godaddy/tartufo/issues/268) - Adds a new
   `--recurse / --no-recurse` flag which allows users to recursively scan the entire directory or just
   the root directory
+* [#256](https://github.com/godaddy/tartufo/issues/256) - Deprecated `--rules` in
+  favor of a new `rule-patterns` config option. This is the final piece of config
+  that was still stored in an external file.
 
 v3.0.0-alpha.1 - 11 November 2021
 ---------------------------------
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -52,6 +52,58 @@ this:
 Note that all options specified in a configuration file are treated as
 defaults, and will be overridden by any options specified on the command line.
 
-For a full list of configuration options, check out the :doc:`usage` document.
+For a full list of available command line options, check out the :doc:`usage`
+document.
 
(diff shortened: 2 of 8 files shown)


tarkatronic:
There was no good place to document a new configuration option like this, so I added it to the end of the `configuration.rst`. Given that I will be doing a bit of a reorganization of documentation, I think this makes the most sense, and I will make an effort to move other similar configuration options into this same space.

sushantmimani (approved):
LGTM 🦅

rbailey-godaddy (approved):
LGTM!

PR #289, opened 2021-12-02
Updating --include-path-patterns and --exclude-path-patterns to use a table of tables

### LINKED ISSUE #257: Change `--include-path-patterns` and `--exclude-path-patterns` to use a table of tables

mayuriesha:

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
closes #257 

## What's new?
- Supports table of tables format for `--include-path-patterns` and `--exclude-path-patterns` with an option to specify the reason.
Ex: include-path-patterns = [{path-pattern="path-pattern", reason="reason to include path patterns"}]
- Warns users about deprecation of old format
- Making sure either one the formats is used until the old format is completely deprecated

### DIFF
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -106,10 +106,11 @@ class TartufoCLI(click.MultiCommand):
     multiple=True,
     hidden=True,
     help="""Specify a regular expression which matches Git object paths to
-    include in the scan. This option can be specified multiple times to include
-    multiple patterns. If not provided (default), all Git object paths are
-    included unless otherwise excluded via the --exclude-path-patterns
-    option.""",
+    include in the scan along with the reason. This option can be specified
+    multiple times to include multiple patterns. If not provided (default),
+    all Git object paths are included unless otherwise excluded via the
+    --include-path=[{path_pattern='path_pattern', reason='Test path pattern
+    that needs to be included'}] option.""",
 )
 @click.option(
     "-x",
@@ -117,10 +118,11 @@ class TartufoCLI(click.MultiCommand):
     multiple=True,
     hidden=True,
     help="""Specify a regular expression which matches Git object paths to
-    exclude from the scan. This option can be specified multiple times to
-    exclude multiple patterns. If not provided (default), no Git object paths
-    are excluded unless effectively excluded via the --include-path-patterns
-    option.""",
+    exclude from the scan along with the reason. This option can be specified
+    multiple times to exclude multiple patterns. If not provided (default),
+    no Git object paths are excluded unless effectively excluded via the
+    --exclude-path=[{path_pattern='path_pattern', reason='Test path pattern
+    that needs to be excluded'}] option.""",
 )
 @click.option(
     "-of",
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -88,8 +88,9 @@ class TartufoCLI(click.MultiCommand):
     help="""Specify a regular expression which matches Git object paths to
     include in the scan. Multiple patterns can be included in the config file using
     include-path-patterns = [{path-pattern="pattern", reason="reason to include pattern},].
-    If not provided (default), all Git object paths are included unless otherwise excluded using
-    exclude-path-patterns""",
+    If not provided (default), all Git object paths
+    are included unless otherwise excluded via the --exclude-path-patterns
+    option.""",
 )
 @click.option(
     "-xp",
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -84,7 +84,7 @@ class TartufoCLI(click.MultiCommand):
     "-ip",
     "--include-path-patterns",
     multiple=True,
-    help="""Specify a regular expression which matches Git object paths to
+    help="""[DEPRECATED] Specify a regular expression which matches Git object paths to
     include in the scan. This option can be specified multiple times to include
     multiple patterns. If not provided (default), all Git object paths are
     included unless otherwise excluded via the --exclude-path-patterns
@@ -94,6 +94,28 @@ class TartufoCLI(click.MultiCommand):
     "-xp",
     "--exclude-path-patterns",
     multiple=True,
+    help="""[DEPRECATED] Specify a regular expression which matches Git object paths to
+    exclude from the scan. This option can be specified multiple times to
+    exclude multiple patterns. If not provided (default), no Git object paths
+    are excluded unless effectively excluded via the --include-path-patterns
+    option.""",
+)
+@click.option(
+    "-i",
+    "--include-path",
+    multiple=True,
+    hidden=True,
+    help="""Specify a regular expression which matches Git object paths to
+    include in the scan. This option can be specified multiple times to include
+    multiple patterns. If not provided (default), all Git object paths are
+    included unless otherwise excluded via the --exclude-path-patterns
+    option.""",
+)
+@click.option(
+    "-x",
+    "--exclude-path",
+    multiple=True,
+    hidden=True,
     help="""Specify a regular expression which matches Git object paths to
     exclude from the scan. This option can be specified multiple times to
     exclude multiple patterns. If not provided (default), no Git object paths
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -84,45 +84,24 @@ class TartufoCLI(click.MultiCommand):
     "-ip",
     "--include-path-patterns",
     multiple=True,
-    help="""[DEPRECATED] Specify a regular expression which matches Git object paths to
-    include in the scan. This option can be specified multiple times to include
-    multiple patterns. If not provided (default), all Git object paths are
-    included unless otherwise excluded via the --exclude-path-patterns
-    option.""",
+    hidden=True,
+    help="""Specify a regular expression which matches Git object paths to
+    include in the scan. Multiple patterns can be included in the config file using
+    include-path-patterns = [{path-pattern="pattern", reason="reason to include pattern},].
+    If not provided (default), all Git object paths are included unless otherwise excluded using
+    exclude-path-patterns""",
 )
 @click.option(
     "-xp",
     "--exclude-path-patterns",
     multiple=True,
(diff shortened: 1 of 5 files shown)


tartufo/scanner.py
```diff
@@ -247,13 +248,38 @@ def included_paths(self) -> List[Pattern]:
...
+                    if patterns
+                    else []
+                )
+                return self._included_paths
+            except TypeError:
+                warnings.warn(
+                    "Old format of --include-path-patterns option and config file setup include-path-patterns "
+                    "= ['inclusion pattern'] has been deprecated and will be removed in a future version. "
+                    "Make sure all the inclusions are set up using new pattern i.e. include-path-patterns = "
+                    "[path-pattern='inclusion pattern',reason='reason for inclusion'] in the config file",
```
```suggestion
                    "[{path-pattern='inclusion pattern',reason='reason for inclusion'},] in the config file",
```

tartufo/scanner.py
```diff
@@ -247,13 +248,38 @@ def included_paths(self) -> List[Pattern]:
...
+                patterns = [
+                    pattern["path-pattern"]
+                    for pattern in cast(List[Dict[str, str]], patterns)
+                ]
+                self._included_paths = (
+                    config.compile_path_rules(set(cast(List[str], patterns)))
+                    if patterns
+                    else []
+                )
+                return self._included_paths
```
tarkatronic: Not that it's really a problem, but this and line 277 really shouldn't be necessary. They're a bit redundant, since line 283 always returns this same value.

tartufo/scanner.py
```diff
@@ -247,13 +248,38 @@ def included_paths(self) -> List[Pattern]:
...
+            except TypeError as exc:
+                raise types.ConfigException(
+                    f"Combination of old and new format of include-path-patterns will not be supported.\n{exc}"
+                )
+        self.logger.debug("Included paths was initialized as: %s", self._included_paths)
```
```suggestion
            self.logger.debug("Included paths was initialized as: %s", self._included_paths)
```
We don't need this to log every single time `included_paths` is accessed 😄

tartufo/scanner.py
```diff
@@ -280,12 +306,37 @@ def excluded_paths(self) -> List[Pattern]:
...
+                    config.compile_path_rules(set(cast(List[str], patterns)))
+                    if patterns
+                    else []
+                )
+                return self._excluded_paths
+            except TypeError as exc:
+                raise types.ConfigException(
+                    f"Combination of old and new format of exclude-path-patterns will not be supported.\n{exc}"
+                )
+        self.logger.debug("Excluded paths was initialized as: %s", self._excluded_paths)
```
tarkatronic: All the same comments apply here as in `included_paths` above.

tartufo/scanner.py
```diff
@@ -692,27 +743,6 @@ def load_repo(self, repo_path: str) -> pygit2.Repository:
...
-            exclude_patterns = list(data.get("exclude_path_patterns", ()))
-            repo_exclude_file = data.get("exclude_paths", None)
-            if repo_exclude_file:
-                repo_exclude_file = pathlib.Path(repo_path, repo_exclude_file)
-                if repo_exclude_file.exists():
-                    with repo_exclude_file.open() as handle:
-                        exclude_patterns += handle.readlines()
-            if exclude_patterns:
-                exclude_patterns = config.compile_path_rules(exclude_patterns)
-                self._excluded_paths = list(set(self.excluded_paths + exclude_patterns))
```
tarkatronic, line 727: With this code block gone, how do these supplemental patterns now get added to the config?

tests/test_git_repo_scanner.py
```diff
@@ -61,24 +61,26 @@ def test_load_repo_does_not_filter_submodules_when_requested(
...
         )
+        self.global_options.include_path_patterns = ("foo/",)
         test_scanner = scanner.GitRepoScanner(
             self.global_options, self.git_options, str(self.data_dir)
         )
         test_scanner.load_repo("../tartufo")
         self.assertCountEqual(
             test_scanner.included_paths,
-            [re.compile("foo/"), re.compile("tartufo/"), re.compile("scripts/")],
+            [re.compile("foo/")],
```
tarkatronic: This test doesn't seem to me like it is proving the same intent anymore. Especially now that you added the `self.global_options.include_path_patterns = ("foo/",)`

Now, this test only shows that we load patterns from global options. The intent of this test is to show that:

* `global_options` include `"foo/"`
* `load_repo` is called with a separate repository, with a config file in it.
* That config file includes `"bar/"`
* The `included_paths` must now contain both `"foo/"`, and `"bar/"`

tests/test_git_repo_scanner.py
```diff
@@ -61,24 +61,26 @@ def test_load_repo_does_not_filter_submodules_when_requested(
...
 
     @mock.patch("pygit2.Repository", new=mock.MagicMock())
     @mock.patch("tartufo.config.load_config_from_path")
     def test_extra_exclusions_get_added(self, mock_load: mock.MagicMock):
         mock_load.return_value = (
             self.data_dir / "pyproject.toml",
-            {"exclude_paths": "exclude-files", "exclude_path_patterns": ("bar/",)},
+            {"exclude_path_patterns": ("bar/",)},
         )
+        self.global_options.exclude_path_patterns = ("bar/",)
```
tarkatronic, line 83: Same comment as the test above.

tarkatronic (changes requested):

tartufo/scanner.py
mayuriesha: I believe we need this return as we need to stop further execution if the user has new style of include and exclude path patterns

tartufo/cli.py
```diff
@@ -92,19 +92,23 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
...
     multiple=True,
+    hidden=True,
     help="""Specify a regular expression which matches Git object paths to
-    include in the scan. This option can be specified multiple times to include
-    multiple patterns. If not provided (default), all Git object paths are
-    included unless otherwise excluded via the --exclude-path-patterns
+    include in the scan. Multiple patterns can be included in the config file using
+    include-path-patterns = [{path-pattern="pattern", reason="reason to include pattern},].
+    If not provided (default), all Git object paths
+    are included unless otherwise excluded via the --exclude-path-patterns
```
rbailey-godaddy, line 100: I'm picking arbitrarily on this one but also consider L109-L111. Can at least one of these mention what happens when a file matches both `include-path-patterns` and `exclude-path-patterns`, i.e. which takes precedence?

rbailey-godaddy: I see while researching other changes below that a file which is both included and excluded will be excluded.

tartufo/scanner.py
rbailey-godaddy: So, a couple of things. We're stuck with butt-ugly code because of the logic, but I think a couple of tactical maneuvers could help us out. These comments are inclusive for the block from L259 through L298.

First, we have a problem in that although we are telling people they must use "all old flavor" or "all new flavor", the code actually enforces "all processed toml files must use the same old or new flavor" which isn't what we say, may not be practicable, and isn't necessary.

To combat this, I think we need to process the `patterns` (global) and `include_patterns` (local) lists independently, and combine them at the end after they have been compiled. Tactically, I would also suggest avoiding reuse of `patterns` when we know the type is changing and use of `set` rather than `list` for the result so we don't have to deduplicate manually (and subsequent queries should be more efficient).

In pseudocode, something more like:

```
self._included_paths: Set[Pattern] = set()
for plist in (self.global_options.include_path_patterns, self.config_data.get("include_path_patterns")):
    if plist:
        try:
            self._included_paths.update(
                config.compile_path_rules({x["path-patterns"] for x in plist})
            )
            continue # New way worked, move on
        except TypeError:
            warnings.warn(blah blah blah, DeprecationWarning)
        # Try the old way
        try:
            self._included_paths.update(
                config.compile_path_rules(set(plist))
            )
        except TypeError as exc:
             raise types.ConfigException(you can't mix both types in the same list)
```
We possibly do some extra regex compiling before deduplicating, but I think the clarity is worth the overhead.

rbailey-godaddy: ...and I'd use the same pattern (excuse the pun) for the exclude processing that comes next.

rbailey-godaddy: ...and yes, this changes the return from `List[Pattern]` to `Set[Pattern]` but I can't see anything in the code base cares, and I would expect set membership testing to be more efficient than list membership testing.

tartufo/scanner.py
rbailey-godaddy: I think L760 is handling this, although I'm fuzzy on why the config_data property and setter were removed just above (or alternatively why they existed previously).

rbailey-godaddy (changes requested):
Pretty much one change to relax unintended restriction that needlessly crosses multiple configuration files.

tartufo/scanner.py
emayuri-godaddy: There is another config_data property at line 244. So I am removing the duplicate one here.

mayuriesha:
@rbailey-godaddy I have created an issue to address the last few review comments as part of another PR after 3.0 is released

rbailey-godaddy (approved):
LGTM (keeping in mind the new backlog story)

tarkatronic (approved):
🎉

PR #290, opened 2021-12-03
Updating exclude signatures to use either one of the formats

mayuriesha:

* [x] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
- Updating exclude signatures to use either old or new format

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,14 +145,10 @@ exclude-signatures = [
   "357841766ca69979608e4e7fb1da2addeafe6d15cade15466c39d903cc0488a8", # tests/data/scan_folder/test.txt
   "743b893e30777d9c4e28d3d341ca4ba1c2541a9c62b497dece75c2a64420e770", # tests/test_folder_scanner.py
   "358a1ae040ab71b3aafeb4318e5d9e318ff1b681c5bf0e99d381584a33ee7833", # tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt
+  "9d1b278fa384c9dc58d607130d97740910309d13b29f13f7ed6348738ea4f32c",
+  "9f675d4d7f43e484e07fb199a1e7ec372ef204631c1f4ec054acc3119fbde4cf",
+  "88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5",
 ]
-
-exclude-findings = [
-  {signature="9d1b278fa384c9dc58d607130d97740910309d13b29f13f7ed6348738ea4f32c", reason="High entropy string added in test_folder_scanner file on 16th November 2021"},
-  {signature="9f675d4d7f43e484e07fb199a1e7ec372ef204631c1f4ec054acc3119fbde4cf", reason="High entropy string added in build_and_publish_docker file on 13th June 2019"},
-  {signature="88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5", reason="High entropy string added in nothing file on 30th December 2016"},
-]
-
 json = false
 regex = true
 repo-path = "."
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,9 +145,9 @@ exclude-signatures = [
   "357841766ca69979608e4e7fb1da2addeafe6d15cade15466c39d903cc0488a8", # tests/data/scan_folder/test.txt
   "743b893e30777d9c4e28d3d341ca4ba1c2541a9c62b497dece75c2a64420e770", # tests/test_folder_scanner.py
   "358a1ae040ab71b3aafeb4318e5d9e318ff1b681c5bf0e99d381584a33ee7833", # tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt
-  "9d1b278fa384c9dc58d607130d97740910309d13b29f13f7ed6348738ea4f32c",
-  "9f675d4d7f43e484e07fb199a1e7ec372ef204631c1f4ec054acc3119fbde4cf",
-  "88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5",
+  "9d1b278fa384c9dc58d607130d97740910309d13b29f13f7ed6348738ea4f32c", # tests/test_folder_scanner.py 16-Nov-2021
+  "9f675d4d7f43e484e07fb199a1e7ec372ef204631c1f4ec054acc3119fbde4cf", # build_and_publish_docker_file 13-Jun-2019
+  "88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5", # nothing file, 30-Dec-2016
 ]
 json = false
 regex = true
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -362,7 +362,7 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
             for signature in signatures:
                 if not isinstance(signature, str):
                     raise types.ConfigException(
-                        "Combination of old and new format of exclude-signatures will not be supported."
+                        "Combination of old and new format of exclude-signatures will is not supported."
                     )
             self._excluded_signatures = tuple(cast(List[str], signatures))
         return self._excluded_signatures
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -360,7 +360,7 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
                         signatures.add(signature["signature"])
                     except KeyError as exc:
                         raise types.ConfigException(
-                            f"Required key signature missing in exclude-signatures"
+                            "Required key signature missing in exclude-signatures"
                         ) from exc
                 elif isinstance(signature, str):
                     deprecated = True
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -19,6 +19,8 @@ from typing import (
     Pattern,
     Set,
     Tuple,
+    cast,
+    Union,
 )
 import warnings
 
@@ -150,7 +152,7 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
     global_options: types.GlobalOptions
     logger: logging.Logger
     _scan_lock: threading.Lock = threading.Lock()
-    _excluded_findings: Tuple[str, ...] = ()
+    _excluded_signatures: Union[Tuple[Dict[str, str], ...], Tuple[str, ...]] = ()
     _config_data: MutableMapping[str, Any] = {}
 
     def __init__(self, options: types.GlobalOptions) -> None:
@@ -339,34 +341,31 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
             return False
         return True
 
-    @property
-    def config_data(self):
-        return self._config_data
-
-    @config_data.setter
-    def config_data(self, data: MutableMapping[str, Any]) -> None:
-        self._config_data = data
-
     @cached_property
-    def excluded_findings(self) -> tuple:
-        configured_signatures = []
-        signatures = self.config_data.get("exclude_signatures", None)
-        if signatures:
-            warnings.warn(
-                "--exclude-signatures has been deprecated and will be removed in a future version. "
-                "Make sure all the exclusions are moved to exclude-findings section with new format. Example: "
-                "exclude-findings = [{signature='signature', reason='The reason of excluding the signature'}]",
-                DeprecationWarning,
-            )
-            configured_signatures.extend(signatures)
-        findings = self.config_data.get("exclude_findings", None)
-        if findings:
-            configured_signatures.extend([finding["signature"] for finding in findings])
-
-        self._excluded_findings = tuple(
-            set(self.global_options.exclude_signatures + tuple(configured_signatures))
-        )
-        return self._excluded_findings
+    def excluded_signatures(self) -> tuple:
+        if len(self._excluded_signatures) == 0:
+            signatures = list(self.global_options.exclude_signatures or ())
+            try:
+                signatures = [
+                    signature["signature"]
+                    for signature in cast(List[Dict[str, str]], signatures)
+                ]
+                self._excluded_signatures = tuple(cast(List[str], signatures))
+                return self._excluded_signatures
+            except TypeError:
+                warnings.warn(
+                    "--exclude-signatures has been deprecated and will be removed in a future version. "
+                    "Make sure all the exclusions are moved to exclude-findings section with new format. Example: "
+                    "exclude-findings = [{signature='signature', reason='The reason of excluding the signature'}]",
+                    DeprecationWarning,
+                )
+            for signature in signatures:
+                if not isinstance(signature, str):
+                    raise types.ConfigException(
+                        "Combination of old and new format of include-path-patterns will not be supported."
+                    )
+            self._excluded_signatures = tuple(cast(List[str], signatures))
+        return self._excluded_signatures
 
     def signature_is_excluded(self, blob: str, file_path: str) -> bool:
         """Find whether the signature of some data has been excluded in configuration.
@@ -376,8 +375,8 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
         """
         return (
             blob
-            in self.excluded_findings  # Signatures themselves pop up as entropy matches
-            or util.generate_signature(blob, file_path) in self.excluded_findings
+            in self.excluded_signatures  # Signatures themselves pop up as entropy matches
+            or util.generate_signature(blob, file_path) in self.excluded_signatures
         )
 
     @staticmethod
@@ -691,7 +690,6 @@ class GitRepoScanner(GitScanner):
         except (FileNotFoundError, types.ConfigException):
             config_file = None
         if config_file and config_file != self.global_options.config:
-            self.config_data = data
             include_patterns = list(data.get("include_path_patterns", ()))
             repo_include_file = data.get("include_paths", None)
             if repo_include_file:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -20,7 +20,6 @@ from typing import (
     Set,
     Tuple,
     cast,
-    Union,
 )
 import warnings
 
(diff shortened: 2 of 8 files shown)


tartufo/scanner.py
```diff
@@ -339,34 +341,31 @@ def should_scan(self, file_path: str):
...
+                    signature["signature"]
+                    for signature in cast(List[Dict[str, str]], signatures)
+                ]
+                self._excluded_signatures = tuple(cast(List[str], signatures))
+                return self._excluded_signatures
+            except TypeError:
+                warnings.warn(
+                    "--exclude-signatures has been deprecated and will be removed in a future version. "
+                    "Make sure all the exclusions are moved to exclude-findings section with new format. Example: "
+                    "exclude-findings = [{signature='signature', reason='The reason of excluding the signature'}]",
```
```suggestion
                    "Configuring exclude-signatures as string has been deprecated and support for this format will "
                    "be removed in the future. Please make sure to update your exclude-signatures configuration to "
                    "an array of tables. For example: exclude-signatures = [{signature='signature', reason='The "
                    "reason of excluding the signature'}]",
```

tartufo/scanner.py
```diff
@@ -339,34 +341,31 @@ def should_scan(self, file_path: str):
...
+                    DeprecationWarning,
+                )
+            for signature in signatures:
+                if not isinstance(signature, str):
+                    raise types.ConfigException(
+                        "Combination of old and new format of include-path-patterns will not be supported."
```
```suggestion
                        "Combination of old and new format of exclude-signatures will not be supported."
```

tartufo/scanner.py
```diff
@@ -691,7 +690,6 @@ def load_repo(self, repo_path: str) -> pygit2.Repository:
         except (FileNotFoundError, types.ConfigException):
             config_file = None
         if config_file and config_file != self.global_options.config:
-            self.config_data = data
```
tarkatronic: We need to be certain that we still load excluded signatures from the target repository's config somehow.

tartufo/scanner.py
```diff
@@ -150,7 +152,7 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
     global_options: types.GlobalOptions
     logger: logging.Logger
     _scan_lock: threading.Lock = threading.Lock()
-    _excluded_findings: Tuple[str, ...] = ()
+    _excluded_signatures: Union[Tuple[Dict[str, str], ...], Tuple[str, ...]] = ()
```
tarkatronic: This should still only ever be `Tuple[str, ...]`, right? This is never the value directly from `global_options`, but only the eventually interpreted value, which would always be a tuple of strings.
```suggestion
    _excluded_signatures: Tuple[str, ...] = ()
```

tests/test_base_scanner.py
```diff
@@ -276,10 +276,7 @@ class SignatureTests(ScannerTestCase):
     def test_matched_signatures_are_excluded(self, mock_signature: mock.MagicMock):
         mock_signature.return_value = "foo"
         test_scanner = TestScanner(self.options)
-        self.options.exclude_signatures = ()
-        test_scanner.config_data = {
-            "exclude_findings": [{"signature": "foo", "reason": "foo finding"}]
-        }
+        self.options.exclude_signatures = ("foo",)
         self.assertTrue(test_scanner.signature_is_excluded("bar", "blah"))
```
tarkatronic, line 280: The new property you created is `excluded_signatures`, but here you are referencing `exclude_signatures`. That makes me think, this should not be passing.

tarkatronic (changes requested):

pyproject.toml
```diff
@@ -145,14 +145,10 @@ exclude-signatures = [
   "357841766ca69979608e4e7fb1da2addeafe6d15cade15466c39d903cc0488a8", # tests/data/scan_folder/test.txt
   "743b893e30777d9c4e28d3d341ca4ba1c2541a9c62b497dece75c2a64420e770", # tests/test_folder_scanner.py
   "358a1ae040ab71b3aafeb4318e5d9e318ff1b681c5bf0e99d381584a33ee7833", # tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt
+  "9d1b278fa384c9dc58d607130d97740910309d13b29f13f7ed6348738ea4f32c",
+  "9f675d4d7f43e484e07fb199a1e7ec372ef204631c1f4ec054acc3119fbde4cf",
+  "88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5",
```
rbailey-godaddy:
```suggestion
  "9d1b278fa384c9dc58d607130d97740910309d13b29f13f7ed6348738ea4f32c", # tests/test_folder_scanner.py 16-Nov-2021
  "9f675d4d7f43e484e07fb199a1e7ec372ef204631c1f4ec054acc3119fbde4cf", # build_and_publish_docker_file 13-Jun-2019
  "88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5", # nothing file, 30-Dec-2016
```
This is just roughly transcribing the reason text from the new style. Frankly it might be even better to just merge in the new exclude-findings stuff (with reason) here and then convert the old lines to the new format, but I could understand not wanting to mess with it at this point. ;-)

tartufo/scanner.py
rbailey-godaddy: "will not" or "is not"?

tartufo/scanner.py
rbailey-godaddy: This is futures material, but before I lose the thought -- might it be reasonable to add a default exclusion that ignores signatures in pyproject.toml?

tartufo/types.py
```diff
@@ -67,8 +66,7 @@ class GlobalOptions:
     include_path_patterns: Tuple[str, ...]
     exclude_path_patterns: Tuple[str, ...]
     exclude_entropy_patterns: Tuple[Dict[str, str], ...]
-    exclude_signatures: Tuple[str, ...]
-    exclude_findings: Tuple[Dict[str, str], ...]
+    exclude_signatures: Union[Tuple[Dict[str, str], ...], Tuple[str, ...]]
```
rbailey-godaddy, line 69: I'm guessing @tarkatronic 's comment above applies here too?

rbailey-godaddy (commented):
A few quick impressions as I look at this first thing on my return... ;-) I'm assuming you'll have a few more commits coming and I'll re-review at that point.

tests/test_base_scanner.py
mayuriesha: Here we are accessing the global options exclude_signatures and the tests are for another method signature_is_excluded

tartufo/types.py
mayuriesha: Exclude signatures will be supporting both strings and dictionary from the config file and so I believe we will need to support both the formats here.

rbailey-godaddy (changes requested):
I think this has the same basic problem as the other PR, in the sense that we want to process the global and local lists separately so they are allowed to have different representations. You should be able to use the same approach to disconnect them.

And, although not strictly necessary, I would advocate changing the return from a tuple to a set (but leave it alone if @tarkatronic or anybody else thinks this would be problematic).

mayuriesha:
@rbailey-godaddy I have created an issue to address the last few review comment as part of another PR after 3.0 is released

tartufo/scanner.py
```diff
@@ -348,25 +349,36 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
-            configured_signatures.extend(signatures)
-        findings = self.config_data.get("exclude_findings", None)
-        if findings:
-            configured_signatures.extend([finding["signature"] for finding in findings])
-
-        self._excluded_findings = tuple(
-            set(self.global_options.exclude_signatures + tuple(configured_signatures))
-        )
-        return self._excluded_findings
+    def excluded_signatures(self) -> tuple:
```
```suggestion
    def excluded_signatures(self) -> Tuple[str, ...]:
```

tartufo/scanner.py
```diff
@@ -150,7 +151,7 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
     global_options: types.GlobalOptions
     logger: logging.Logger
     _scan_lock: threading.Lock = threading.Lock()
-    _excluded_findings: Tuple[str, ...] = ()
+    _excluded_signatures: Tuple[str, ...] = ()
```
```suggestion
    _excluded_signatures: Optional[Tuple[str, ...]] = None
```

tartufo/scanner.py
```diff
@@ -348,25 +349,36 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
-        findings = self.config_data.get("exclude_findings", None)
-        if findings:
-            configured_signatures.extend([finding["signature"] for finding in findings])
-
-        self._excluded_findings = tuple(
-            set(self.global_options.exclude_signatures + tuple(configured_signatures))
-        )
-        return self._excluded_findings
+    def excluded_signatures(self) -> tuple:
+        if len(self._excluded_signatures) == 0:
```
```suggestion
        if self._excluded_signatures is None:
```
This way, if no exclusions are configured, we don't re-process every time. Before calculation this will be `None`; after calculation it will be `()`, and never re-calculated.

tartufo/scanner.py
```diff
@@ -348,25 +349,36 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
+            signatures = list(self.global_options.exclude_signatures or ())
+            exclude_signatures = list(self.config_data.get("exclude_signatures", ()))
```
```suggestion
            signatures += list(self.config_data.get("exclude_signatures", ()))
```
Gather these all into one list so that only have to iterate over the once, and in a single format

tests/test_git_repo_scanner.py
```diff
@@ -98,18 +98,14 @@ def test_extra_exclusions_get_added(self, mock_load: mock.MagicMock):
     def test_extra_signatures_get_added(self, mock_load: mock.MagicMock):
         mock_load.return_value = (
             self.data_dir / "pyproject.toml",
-            {
-                "exclude_signatures": ["foo", "bar"],
-                "exclude_findings": [{"signature": "foo/bar", "reason": "test reason"}],
-            },
+            {"exclude_signatures": ["foo", "bar"]},
```
tarkatronic:
```suggestion
            {"exclude_signatures": ["foo"]},
```

tests/test_git_repo_scanner.py
```diff
@@ -98,18 +98,14 @@ def test_extra_exclusions_get_added(self, mock_load: mock.MagicMock):
...
         )
+        self.global_options.exclude_signatures = ("foo", "bar")
```
tarkatronic:
```suggestion
        self.global_options.exclude_signatures = ("bar")
```
This way we pull just one of the exclusions from each source, and ensure they are being joined together.

tests/test_git_repo_scanner.py
```diff
@@ -98,18 +98,14 @@ def test_extra_exclusions_get_added(self, mock_load: mock.MagicMock):
...
         test_scanner = scanner.GitRepoScanner(
             self.global_options, self.git_options, str(self.data_dir)
         )
         test_scanner.load_repo("../tartufo")
-        self.assertEqual(
-            sorted(test_scanner.excluded_findings), ["bar", "foo", "foo/bar"]
-        )
+        self.assertEqual(sorted(test_scanner.excluded_signatures), ["bar", "foo"])
```
tarkatronic:
```suggestion
        self.assertCountEqual(test_scanner.excluded_signatures, ["bar", "foo"])
```
Handy trick for comparing lists in unit tests: This makes sure that all of the elements match, but doesn't care about the order of those elements.

tarkatronic (changes requested):

tartufo/scanner.py
```diff
@@ -348,25 +349,36 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
+            try:
+                signatures = signatures + [
+                    x for x in exclude_signatures if x not in signatures
+                ]
```
rbailey-godaddy: This is unnecessary if you make the suggested change in L355.

tartufo/scanner.py
```diff
@@ -348,25 +349,36 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
+                signatures = [
+                    signature["signature"]
+                    for signature in cast(List[Dict[str, str]], signatures)
+                ]
+                self._excluded_signatures = tuple(cast(List[str], signatures))
```
rbailey-godaddy: ...at which point:
```suggestion
                self._excluded_signatures = (signature["signature"] for signature in signatures)
```

tartufo/scanner.py
```diff
@@ -348,25 +349,36 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
+                warnings.warn(
+                    "Configuring exclude-signatures as string has been deprecated and support for this format will "
+                    "be removed in the future. Please make sure to update your exclude-signatures configuration to "
+                    "an array of tables. For example: exclude-signatures = [{signature='signature', reason='The "
+                    "reason of excluding the signature'}]",
+                    DeprecationWarning,
+                )
+            try:
+                signatures = list(set(signatures + exclude_signatures))
+                self._excluded_signatures = tuple(cast(List[str], signatures))
```
rbailey-godaddy: And again, we've already merged the lists and eliminated duplicates, so this can be simplified

```suggestion
                self._excluded_signatures = tuple(signatures)
```

tartufo/scanner.py
```diff
@@ -348,25 +349,36 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
+            except TypeError as exc:
+                raise types.ConfigException(
+                    f"Combination of old and new format of exclude-signatures is not supported.\n{exc}"
+                )
```
rbailey-godaddy: As it stands now, this is never going to be triggered because the only protected code is a simple conversion from a set to a tuple (which is insensitive to the type(s) of the members). I believe it would be better to raise this error here than weird errors later when the members turn out to be weird (due to mixing and matching) -- but that probably means you need to explicitly look at each member and raise this exception `if not isinstance(signature, str)`.

rbailey-godaddy (changes requested):
This layers on top of @tarkatronic 's comments, mostly in eliminating dangling use of removed variables and preserving exception semantics.

tartufo/scanner.py
tarkatronic: We may still need the `cast()` here; that was an appeasement for `mypy`

tarkatronic (approved):

tartufo/scanner.py
```diff
@@ -349,19 +349,17 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
-                ]
-                signatures = [
+                signatures = tuple(
                     signature["signature"]
-                    for signature in cast(List[Dict[str, str]], signatures)
-                ]
-                self._excluded_signatures = tuple(cast(List[str], signatures))
+                    for signature in cast(Tuple[Dict[str, str]], signatures)
+                )
+                self._excluded_signatures = tuple(set(signatures))
```
rbailey-godaddy: This makes no sense to me. Why use a tuple comprehension, and then convert it to a set, and then convert it back to a tuple -- instead of starting with a set comprehension and then converting it once? If we're worried about mutating the type of `signatures`, then just eliminate the intermediate variable and store directly into the destination.

rbailey-godaddy (approved):
I'm lukewarm on the unnecessary type thrashing but okay with this as-is if you just want to get it out and quit fiddling around. ;-)

tartufo/scanner.py
tarkatronic: Yeah... I had very similar thoughts. And had your words echoing in my head as I did it. I just didn't want to have to think about all the other places that might be depending on a tuple vs a set. In theory it shouldn't matter. But... you never know.

tarkatronic: @rbailey-godaddy Since it's already getting to dinner time, I'm happy to leave this stewing overnight in case you want to take a moment in the morning to un-thrashify this code. 😄

tarkatronic: But, also want to keep in mind that we're more than a week overdue for 3.0 now, so... limiting scope is good.

rbailey-godaddy: I pushed a commit that makes me moderately happy. (Well, two because pylint is so anal.) For "proper" usage, this makes no changes that are visible outside of the modified property function. Changes in behavior:

* If you mix old and new style specifications, it now works. (Fixed unit test to confirm this.) This does not affect whether or not you can -- or should -- actually mix styles in your configuration files, only that we aren't artificially restricting things ourselves.
* If you pass new-style entry that is missing signature key, it tells you that explicitly. Previously it would raise an unhandled KeyError.
* If you (somehow) pass an entry that's neither old-style nor new-style, it tells you that explicitly. Previously such alien input would be treated as "old-style" and inserted directly into the return value tuple, where it probably would be ignored.

As a matter of taste, "tells you explicitly" means "raises meaningful exception and aborts" rather than "whines, ignores the erroneous entry, and continues." I feel if you have b0rked configuration files, you should fix them.

Also, I marked all of the conversations above "resolved" where they discussed outdated code.

tartufo/scanner.py
```diff
@@ -351,30 +350,34 @@ def config_data(self, data: MutableMapping[str, Any]) -> None:
...
-                    for signature in cast(Tuple[Dict[str, str]], signatures)
-                )
-                self._excluded_signatures = tuple(set(signatures))
-                return self._excluded_signatures
-            except TypeError:
+            signatures: Set[str] = set()
+            deprecated = False
+            for signature in tuple(
+                self.global_options.exclude_signatures or []
+            ) + tuple(self.config_data.get("exclude_signatures", [])):
```
tarkatronic, line 357:
```pycon
>>> () + []
TypeError: can only concatenate tuple (not "list") to tuple
>>>
```
Careful of the mixture of lists and tuples here. That was part of the cause for the previous type thrashing. 😄

rbailey-godaddy: That's why both of them are being explicitly converted from lists to tuples here. Also, I played with this extensively last night and python definitely thinks both `self.global_options.exclude_signatures` and `self.config_data["exclude_signatures"]` are lists, which is why I used empty lists for defaults -- to maintain type consistency.

This is strangely bizarre in the sense that `self.global_options.exclude_signatures` pretty clearly appears to be defined as a tuple rather than a list, but if you try removing either of the tuple conversions you get errors complaining that they are lists and that you aren't allowed to concatenate lists this way (or combine a list and a tuple, as the case may be).

I didn't have enough ambition to trace down where the actual content is coming from to understand why mypy isn't whining about some assignment somewhere that is putting a list into a tuple type... Unless it's happening under the covers in click.

tarkatronic (approved):

PR #297, opened 2021-12-09
Better docker build

tarkatronic:
While our docker image was building now, it wasn't actually working, because installing `libgit2-dev` did not install `libgit2`. So now we both build the Docker image and test it in our CI pipeline.


sushantmimani (approved):
LGTM 🦅

PR #319, opened 2022-01-11
Entropy scanning fixups

### LINKED ISSUE #315: Entropy scanning producing extra (incorrect) outputs
## 🐛 Bug Report

Values like below are being flagged as high entropy findings when they shouldn't.

```
policy/service-role/AWSLambdaVPCAccessExecutionRole
b/Standards-Best-Practices/Speed-of-Delivery/TEMPLATE
```

## Expected Behavior

Values should not be flagged as high entropy findings.

## Environment

tartufo v3.0.0

rbailey-godaddy:

* [ ] Tests (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

"Almost" -- see description below for corner cases.

## Issue Linking

closes #315

## What's new?
The loosey-goosey "base64-anything" matching introduced in 3.0.0 turned out to be problematic because we several examples of things that looked like base64 encodings but weren't and which generated new issues thanks to the extended combined alphabet. Simple tweaks looked problematic (read "nondeterministic") going forward.

This fix is a little bit more invasive.

* We scan for BASE64 and BASE64URL encodings separately (using the same flakey expressions as pre-3.0.0 did) so they cannot bleed into each other.  The cost of this is that we now make 3 (vs 2) regex scans of input text. A number of unit tests required revision to account for the additional calls.
* To offset this cost, we further adjust the regex expressions to match a minimum of 20 characters, offloading much of the work from function `util.find_strings_by_regex()` which spent most of its time throwing away short strings. Nothing in the tartufo codebase allows users to supply a different minimum length; I added a warning for folks who might be building other code on top of tartufo.
* Additionally, `scanner.scan_entropy()` used to split lines into words and scan the words individually. Since the supplied regexs will do this by themselves anyway, we eliminate the `line.split()` step and eliminate a pass over the input text. This required fixes to a number of unit tests where we generally make fewer calls on longer targets.
* But wait! There's more! Many strings are both base64 and base64url, and we don't want to generate duplicate (or overlapping) issues. That resulted in the following two changes...
* `scanner.evaluate_entropy_string()` now returns an optional issue instead of being a generator (that returned either 1 or 0 issues). This allows its only caller to inspect the return before passing it back to higher-level callers.
* `scanner.scan_entropy()` now jumps through deduplication hoops:
  - results from both base64 and base64url scans are combined in a set to eliminate strict duplicates; as a side effect, it is possible that lines with multiple issues (in the same line) may return the issues in a different order.
  - target strings are evaluated from longest to shortest, and we no longer report an issue for a substring if we already reported an issue on the longer string containing it. We do not strictly test that the substring is located within the longer string, so a line that contained "bad-string even-more-bad-string" would not generate an issue for the initial "bad-string" even though it appears before the subsequent "even-more-bad-string" (which would get an issue).
  - We also test for duplicate hex encoding issues (because a hex string is also a base64 string) -- this could not happen with default sensitivity (backstory omitted) but is nice to close off.

My testing shows variations in the noise level when tartufo scans itself, but it seems like the extra overhead should be approximately balanced by the efficiency gains and we'll see a wash. Time will tell.

Note it is *still* possible to get new issues (not reported by 2.x) if a string begins or ends with `-` or `_` and the old string didn't generate an issue but the new longer string does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+v3.0.x - TBD
+------------
+
+Bug fixes:
+
+* [315](https://github.com/godaddy/tartufo/issues/315) - Rework entropy scanning
+  to reduce number of new issues generated for text that passed 2.x scans without
+  issues. Some corner cases may remain that will require exclusions to silence.
+  See [319](https://github.com/godaddy/tartufo/pull/319) for details.
+
 v3.0.0 - 5 January 2022
 -----------------------
 
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -36,8 +36,9 @@ from tartufo.types import (
     Scope,
 )
 
-BASE64_REGEX = re.compile(r"[A-Z0-9=+/_-]+", re.IGNORECASE)
-HEX_REGEX = re.compile(r"[0-9A-F]+", re.IGNORECASE)
+BASE64_REGEX = re.compile(r"[A-Z0-9=+/]{20,}", re.IGNORECASE)
+BASE64URL_REGEX = re.compile(r"[A-Z0-9=_-]{20,}", re.IGNORECASE)
+HEX_REGEX = re.compile(r"[0-9A-F]{20,}", re.IGNORECASE)
 
 
 class Issue:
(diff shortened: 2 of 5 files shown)


rbailey-godaddy:
I'm closing this PR; it's been sitting without action for over a year, and clearly:
* users apparently aren't too upset about the existing behavior
* we really don't need another (gratuitious) incompatible finding change

PR #328, opened 2022-02-10
Eliminate OOM errors in tartufo by using temporary file for issues

jgowdy:

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
- This pull request introduces new code that buffers excessive issues to a temporary file to avoid exhausting memory with a massive list[Issue] object.  This avoids out of memory errors for repos with very high issue counts and makes the memory usage of tartufo during a scan much more rational and predictable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+Unreleased
+----------
+
+Features:
+
+* [#328](https://github.com/godaddy/tartufo/pull/328) - Buffer issues to temporary file as needed
+
 v3.0.0 - 5 January 2022
 -----------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Unreleased
 
 Features:
 
-* [#328](https://github.com/godaddy/tartufo/pull/328) - Buffer issues to temporary file as needed
+* [#328](https://github.com/godaddy/tartufo/pull/328) - Buffer issues beyond --buffer-size to a temporary file
 
 v3.0.0 - 5 January 2022
 -----------------------
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -183,7 +183,7 @@ class TartufoCLI(click.MultiCommand):
     help="If specified, temporary files will be written to the specified path",
 )
 @click.option(
-    "--max-buffered-issues",
+    "--buffer-size",
     type=int,
     default=10000,
     show_default=True,
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -170,6 +170,26 @@ class TartufoCLI(click.MultiCommand):
     "the results of individual runs of tartufo separated.",
 )
 @click.option(
+    "-td",
+    "--temp-dir",
+    type=click.Path(
+        exists=False,
+        file_okay=False,
+        dir_okay=True,
+        writable=True,
+        resolve_path=True,
+        allow_dash=False,
+    ),
+    help="If specified, temporary files will be written to the specified path",
+)
+@click.option(
+    "--max-buffered-issues",
+    type=int,
+    default=10000,
+    show_default=True,
+    help="Maximum number of issue to buffer in memory before shifting to temporary file buffering",
+)
+@click.option(
     "--git-rules-repo",
     help="A file path, or git URL, pointing to a git repository containing regex "
     "rules to be used for scanning. By default, all .json files will be loaded "
@@ -304,7 +324,7 @@ def process_exit(
     scan: scanner.ScannerBase,
     **_kwargs: config.OptionTypes,
 ):
-    if scan.issues:
+    if scan.issue_count > 0:
         ctx.exit(1)
 
     ctx.exit(0)
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -43,7 +43,6 @@ from tartufo.types import (
 
 BASE64_REGEX = re.compile(r"[A-Z0-9=+/_-]+", re.IGNORECASE)
 HEX_REGEX = re.compile(r"[0-9A-F]+", re.IGNORECASE)
-OUTPUT_SEPARATOR = "~~~~~~~~~~~~~~~~~~~~~"
 
 
 class Issue:
@@ -60,6 +59,7 @@ class Issue:
     issue_type: types.IssueType
     issue_detail: Optional[str]
     matched_string: str
+    OUTPUT_SEPARATOR: str = "~~~~~~~~~~~~~~~~~~~~~"
 
     def __init__(
         self, issue_type: types.IssueType, matched_string: str, chunk: types.Chunk
@@ -106,7 +106,7 @@ class Issue:
         diff_body = diff_body.replace(
             self.matched_string, util.style_warning(self.matched_string)
         )
-        output.append(OUTPUT_SEPARATOR)
+        output.append(self.OUTPUT_SEPARATOR)
         output.append(util.style_ok(f"Reason: {self.issue_type.value}"))  # type: ignore
         if self.issue_detail:
             output.append(util.style_ok(f"Detail: {self.issue_detail}"))
@@ -118,7 +118,7 @@ class Issue:
         ]
 
         output.append(diff_body)
-        output.append(OUTPUT_SEPARATOR)
+        output.append(self.OUTPUT_SEPARATOR)
         return "\n".join(output)
 
     def __bytes__(self) -> bytes:
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -510,8 +510,7 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
             try:
                 length = int.from_bytes(self.issue_file.read(4), "little")
                 buf = self.issue_file.read(length)
-                issue = pickle.loads(gzip.decompress(buf))
-                yield from issue
+                yield from pickle.loads(gzip.decompress(buf))
             except EOFError:
                 self.logger.debug("pickle.load raised EOFError, exiting")
                 yield from self._issue_list
--- a/tartufo/scanner.py
+++ b/tartufo/scanner.py
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
 # -*- coding: utf-8 -*-
 
 import abc
@@ -9,6 +10,9 @@ import math
 import pathlib
 import re
 import threading
+import tempfile
+import pickle
+import gzip
 from typing import (
     Any,
     Dict,
@@ -19,6 +23,7 @@ from typing import (
     Pattern,
     Set,
     Tuple,
+    IO,
 )
 import warnings
 
@@ -38,26 +43,23 @@ from tartufo.types import (
 
 BASE64_REGEX = re.compile(r"[A-Z0-9=+/_-]+", re.IGNORECASE)
 HEX_REGEX = re.compile(r"[0-9A-F]+", re.IGNORECASE)
+OUTPUT_SEPARATOR = "~~~~~~~~~~~~~~~~~~~~~"
 
 
 class Issue:
     """Represent an issue found while scanning a target."""
 
     __slots__ = (
-        "OUTPUT_SEPARATOR",
         "chunk",
         "issue_type",
         "issue_detail",
         "matched_string",
-        "logger",
     )
 
-    OUTPUT_SEPARATOR: str
     chunk: types.Chunk
     issue_type: types.IssueType
     issue_detail: Optional[str]
     matched_string: str
-    logger: logging.Logger
 
     def __init__(
         self, issue_type: types.IssueType, matched_string: str, chunk: types.Chunk
@@ -67,13 +69,10 @@ class Issue:
         :param matched_string: The string that was identified as a potential issue
         :param chunk: The chunk of data where the match was found
         """
-        self.OUTPUT_SEPARATOR = "~~~~~~~~~~~~~~~~~~~~~"  # pylint: disable=invalid-name
         self.issue_detail = None
-
         self.issue_type = issue_type
         self.matched_string = matched_string
         self.chunk = chunk
-        self.logger = logging.getLogger(__name__)
 
     def as_dict(self, compact=False) -> Dict[str, Optional[str]]:
         """Return a dictionary representation of an issue.
@@ -107,7 +106,7 @@ class Issue:
         diff_body = diff_body.replace(
             self.matched_string, util.style_warning(self.matched_string)
         )
-        output.append(self.OUTPUT_SEPARATOR)
+        output.append(OUTPUT_SEPARATOR)
         output.append(util.style_ok(f"Reason: {self.issue_type.value}"))  # type: ignore
         if self.issue_detail:
             output.append(util.style_ok(f"Detail: {self.issue_detail}"))
@@ -119,7 +118,7 @@ class Issue:
         ]
 
         output.append(diff_body)
-        output.append(self.OUTPUT_SEPARATOR)
+        output.append(OUTPUT_SEPARATOR)
         return "\n".join(output)
 
     def __bytes__(self) -> bytes:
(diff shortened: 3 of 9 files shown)


rbailey-godaddy (changes requested):
I salute your ingenuity in having a repository that (unfortunately) has this many issues in it, and providing an approach for dealing with it. I have a few random questions about it, in no particular order:

* I presume you now take `self._scan_lock` in `issues()` because you are modifying the state of `self._issue_file` and do not want to race with any other threads. Am I correct in recalling that we perform all operations under this lock "start to finish" so we worry only about racing, not changing state (i.e. before taking the lock vs after releasing it)?
* Could you make that 10,000 issue cutoff a variable (say, to `__init__()`, with a default of 10000?) That would make it easier tune for memory sizes and probably easier to unit test. I am undecided about whether it needs to be exposed at the command line (so it's actually tunable) by average users.
* Maybe because it's Monday morning, I could not understand why you commented out `test_rule_patterns_are_read()`

jgowdy:
Yes, there seemed to be usage of that lock to prevent scan() and the property issues from competing with one another.  Holding that lock seemed appropriate given that we were shifting between holding the results in either the list or the file.

You are absolutely right that 10,000 should be a parameter rather than a magic number.  I will make that change.

As far as the test, I would like to fix and restore it.  Unfortunately many of the tests in Tartufo are brittle and violate OOP principles by assuming side effects and modifying protected fields.  It makes any modification or improvement of Tartufo difficult as you always end up breaking tests even if you keep the same signatures when improving the code.  I'll see what I can do to fix the test.

I also need to add a parameter for the temp file itself to allow the user to specify where to place the temp file as it can be quite large and operating systems like MacOS will cut it off by quota even if sufficient disk space is available.

jgowdy:
btw, I don't have a repo that has this many issues, I'm just helping the people who do 😄

jgowdy:
The number of items to buffer in memory and the location of the temporary files have been parameterized.  The previously removed / commented test has been fixed (it had a bug in the test itself).  I've tested this on behalf of a team with a repo that unfortunately has 5,000,000 issues, and the scan does seem to complete.  While this number of issues may seem insurmountable, arguably that's a user decision.  Hypothetically a team could take JSON output and automate the removal or add-signatures needed to remediate the issues.   The tool should function properly regardless of size of repo or number of issues.

rfox-godaddy (approved):
lgtm.  it'd be cool to see benchmarks of this with slow disk storage, but slow is better than oom regardless.

jgowdy:
With regards to disk performance, since it's doing buffered I/O and only writing to disk in batches of tens of thousands of Issue objects we seem to be spending very little time on disk I/O.  There is some additional overhead for the gzip of the pickle data but it is worth doing in my testing.

tartufo/cli.py
```diff
@@ -169,6 +169,33 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
...
+        file_okay=False,
+        dir_okay=True,
+        writable=True,
+        resolve_path=True,
+        allow_dash=False,
+    ),
+    help="If specified, temporary files will be written to the specified path",
+)
+@click.option(
+    "--max-buffered-issues",
```
tarkatronic: What about `--buffer-size`? Does that still get the point across? Just seems easier to not mis-type. 😄

jgowdy:
I ran this under cProfiler with --count-only to suppress writing to stdout and this was the result.
[profiler.txt](https://github.com/godaddy/tartufo/files/8074115/profiler.txt)
 
Heavy functions seem to be:

   157023    5.705    0.000  421.044    0.003 scanner.py:838(chunks)
  4983324   32.828    0.000  295.578    0.000 scanner.py:635(scan_entropy)
   338631    1.107    0.000  262.175    0.001 scanner.py:678(scan_regex)
  6123858  260.871    0.000  260.871    0.000 {method 'findall' of 're.Pattern' objects}
   430900    0.394    0.000  258.653    0.001 repository.py:482(diff)
   430921  256.876    0.001  256.876    0.001 {method 'diff_to_tree' of '_pygit2.Tree' objects}
196892874  176.337    0.000  230.903    0.000 util.py:205(find_strings_by_regex)
   177970  130.625    0.001  130.665    0.001 scanner.py:723(_iter_diff_index)
 15220062    6.295    0.000   26.798    0.000 scanner.py:655(evaluate_entropy_string)
    20927   24.823    0.001   24.823    0.001 {method 'find_similar' of '_pygit2.Diff' objects}
186499114   24.092    0.000   24.092    0.000 {method 'finditer' of 're.Pattern' objects}
465999910   19.425    0.000   19.425    0.000 {method 'group' of 're.Match' objects}
 10575369   11.968    0.000   12.625    0.000 scanner.py:455(signature_is_excluded)
466598669/466597978   11.071    0.000   11.071    0.000 {built-in method builtins.len}
   571983    2.852    0.000    6.316    0.000 scanner.py:513(calculate_entropy)
 23063979    5.049    0.000    5.049    0.000 {method 'split' of 'str' objects}

tartufo/scanner.py
```diff
@@ -149,13 +147,22 @@ class ScannerBase(abc.ABC):  # pylint: disable=too-many-instance-attributes
...
         """
         :param options: A set of options to control the behavior of the scanner
         """
         self.global_options = options
         self.logger = logging.getLogger(__name__)
+        self._issue_file = (
+            tempfile.NamedTemporaryFile(  # pylint: disable=consider-using-with
+                dir=options.temp_dir
+            )
+        )
```
tarkatronic: I wonder if this should be accessed through a property, that way it's only created when necessary. If a repo only has 100 issues, there's no reason to have the added churn of messing with a temporary file. Something like

```python
@property
def issue_file(self) -> IO:
    if not self._issue_file:
        self._issue_file = (
            tempfile.NamedTemporaryFile(  # pylint: disable=consider-using-with
                dir=options.temp_dir
            )
        )
    return self._issue_file
```

tartufo/scanner.py
```diff
@@ -235,9 +246,32 @@ def issues(self) -> List[Issue]:
...
+            self.logger.debug("Scan complete")
 
-        return self._issues
+        if not self._pickling:
+            return self._issue_list
+
+        issues: List[Issue]
+        issues = []
+        count: int
+        count = 0
```
```suggestion
        issues: List[Issue] = []
        count: int = 0
```

tartufo/scanner.py
```diff
@@ -235,9 +246,32 @@ def issues(self) -> List[Issue]:
...
+
+        # Rewind the issue_file
+        self._issue_file.seek(0)
```
tarkatronic: If using the property as suggested above, this (and other references here) would change to `self.issue_file`

tartufo/scanner.py
```diff
@@ -499,7 +533,9 @@ def calculate_entropy(self, data: str) -> float:
             entropy += -probability * math.log2(probability)
         return entropy
 
-    def scan(self) -> Generator[Issue, None, None]:
+    def scan(  # pylint: disable=too-many-statements
```
tarkatronic: I'm generally not comfortable disabling checks like this. When we get into this territory, it makes me think this function is going to be especially hard to test. We may need to look at starting to split it up some more.

jgowdy:
I cleaned up this PR a bit, removing some unnecessary changes, fixing tests, lazily creating the temp file, breaking down the excessively large scan function further, and eliminating code duplication.  I also fixed the fact that we were calling .issues when writing output files which defeated the purpose of the whole exercise.  It should be g2g now.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+Unreleased
+----------
+
+Features:
+
+* [#328](https://github.com/godaddy/tartufo/pull/328) - Buffer issues to temporary file as needed
```
tarkatronic: Let's add some mention in here of the command line options that were added.

tartufo/scanner.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
```
tarkatronic: Fascinating that we're finally hitting this. Makes me think we should make a change (not necessarily in here) that has been bugging me: Moving the `Issue` class out to `types.py`, where it belongs a bit better.

tartufo/scanner.py
```diff
@@ -499,7 +492,44 @@ def calculate_entropy(self, data: str) -> float:
...
+    def load_issues(self) -> Generator[Issue, None, None]:
+        # Rewind the issue_file
+        self.logger.debug("Rewinding pickle file")
+        self.issue_file.seek(0)
+        while True:
+            try:
+                length = int.from_bytes(self.issue_file.read(4), "little")
+                buf = self.issue_file.read(length)
+                issue = pickle.loads(gzip.decompress(buf))
+                yield from issue
```
tarkatronic: Can we rename this to `issues` or `issue_list`? Up until I read how `store_issue` is working, I was assuming this was an individual issue, and wondering why in the world we would `yield from` a single item instead of just `yield`ing it.

tartufo/scanner.py
```diff
@@ -520,9 +550,11 @@ def scan(self) -> Generator[Issue, None, None]:
         # expectation that the write to _completed at the bottom of the critical
         # section cannot be reordered to appear after the implicit release of
         # _scan_lock (when viewed from a competing thread).
+        self.logger.debug("Waiting for scan lock")
         with self._scan_lock:
             if self._completed:
-                yield from self._issues
+                self.logger.debug("Scan already completed")
+                yield from self.load_issues()
```
tarkatronic, line 555: What about cases where the issues are not buffered to file?

tarkatronic: Also, if we have a max buffer size of 10,000 but then we encounter 10,050 issues... the first 10,000 get pushed into the file, but what happens to those remaining 50? I can't see how those ones are ever loaded and reported.

tartufo/scanner.py
```diff
@@ -535,21 +567,20 @@ def scan(self) -> Generator[Issue, None, None]:
                 )
 
             self.logger.info("Starting scan...")
-            self._issues = []
-            for chunk in self.chunks:
+            self._issue_count = 0
+            for chunk in self.chunks:  # pylint: disable=too-many-nested-blocks
```
tarkatronic, line 569: Is this disable still necessary, with the refactoring you've done?

tests/test_base_scanner.py
```diff
@@ -112,13 +114,6 @@ def test_empty_issue_list_causes_scan(self, mock_scan: mock.MagicMock):
         list(test_scanner.issues)  # pylint: disable=pointless-statement
         mock_scan.assert_called()
 
-    @mock.patch("tartufo.scanner.ScannerBase.scan")
-    def test_scanner_does_not_rescan(self, mock_scan: mock.MagicMock):
-        test_scanner = TestScanner(self.options)
-        test_scanner._completed = True  # pylint: disable=protected-access
-        test_scanner.issues  # pylint: disable=pointless-statement
-        mock_scan.assert_not_called()
-
```
tarkatronic, line 121: Do we no longer have this guarantee? If we do, can we get this test added back? If not... why not?

tests/test_cli.py
```diff
@@ -204,6 +204,7 @@ def test_command_exits_with_positive_return_code_when_issues_are_found(
                 types.IssueType.Entropy, "foo", types.Chunk("foo", "/bar", {})
             )
         ]
+        mock_scanner.return_value._issue_count = 1  # pylint: disable=protected-access
```
tarkatronic: This could be changed to use the non-protected attribute like the others in this file

tests/test_cli.py
```diff
@@ -307,6 +315,7 @@ def test_excess_verbosity_also_logs_the_logger_name(
     @mock.patch("tartufo.commands.scan_local_repo.GitRepoScanner")
     def test_excess_verbosity_does_not_exceed_debug(self, mock_scanner: mock.MagicMock):
         mock_scanner.return_value.issues = []
+        mock_scanner.return_value._issue_count = 0  # pylint: disable=protected-access
```
tarkatronic: Same on the non-protected here

tarkatronic (changes requested):
This is a great change @jgowdy; thanks for working on this! This will make for a huge usability enhancement for a lot of people.

I would like to see a few small updates before we get this merged in. Aside from the inline noted changes, the one thing I would mention is that we should be sure to have docstrings on all of our functions.

tartufo/scanner.py
jgowdy: If you see the end of load_issues, after it finishes the file, it yields from _issue_list:

`            except EOFError:
                self.logger.debug("pickle.load raised EOFError, exiting")
                yield from self._issue_list
                return`

tests/test_base_scanner.py
jgowdy: This test is literally testing whether a boolean is evaluated correctly, and it does so by making an assumption that when you call issues, the scan() function is not called.  Yet the scan() function itself guarantees that it won't re-scan when already completed.  This is a poor test that makes assumption about the behavior of the implementation to validate that a protected boolean field is honored.

tartufo/cli.py
jgowdy: I like that

tartufo/scanner.py
jgowdy: I strongly agree with moving Issue to types.py

tests/test_base_scanner.py
tarkatronic: Yeah I agree that the state the test has evolved to is not great. Sounds like we'll need to add another test (external to this PR) to test that functionality in `scan()`, if one doesn't exist already.

rbailey-godaddy: I would have phrased this slightly differently... "Yet the scan() function itself guarantees that it won't re-scan when already completed." True. The point of the test is to verify that guarantee is being honored.

jgowdy: I think the # pylint: disable=protected-access should be a code smell that this test is excessively coupled with the implementation.  This is effectively sub-unit testing.

tartufo/scanner.py
```diff
@@ -38,26 +43,23 @@
 
 BASE64_REGEX = re.compile(r"[A-Z0-9=+/_-]+", re.IGNORECASE)
 HEX_REGEX = re.compile(r"[0-9A-F]+", re.IGNORECASE)
+OUTPUT_SEPARATOR = "~~~~~~~~~~~~~~~~~~~~~"
```
rbailey-godaddy: This seems slightly flaky -- I guess the same complaint could be made about the two REGEX expressions too -- in the sense that these definitions are integral to the behavior of the class but defined outside it and silently consumed.

I presume you moved OUTPUT_SEPARATOR out to reduce the size of each object, but is it feasible to make these class variables? This would be better encapsulation, but if the actual python implementation still ends up duplicating the boilerplate in every instance, that's obviously not a win.

tartufo/scanner.py
rbailey-godaddy: or just combine the two lines as `yield from pickle.loads(gzip.decompress(buf))` given that we don't do anything else with the variable...

rbailey-godaddy (approved):
This is pretty much bikeshedding at this point, so if you want to run with what you have now, I'm okay with it. Thanks again for the work on this one.

tartufo/scanner.py
rbailey-godaddy: I tried a very simple test, and it appears that class variables do not appear in the output of pickled instances. FWIW.

jgowdy: > This seems slightly flaky -- I guess the same complaint could be made about the two REGEX expressions too -- in the sense that these definitions are integral to the behavior of the class but defined outside it and silently consumed.
> 
> 
> 
> I presume you moved OUTPUT_SEPARATOR out to reduce the size of each object, but is it feasible to make these class variables? This would be better encapsulation, but if the actual python implementation still ends up duplicating the boilerplate in every instance, that's obviously not a win.

This is absolutely moved out to reduce the size of the Issue object (both serialized and on the heap), but I think it's totally valid feedback that it should be static class variables rather than global.  I need to learn the Python syntax for that and make the change.  As long as it doesn't end up increasing the size of the serialized object, I'll make this change.  Thanks!

tarkatronic (approved):
Love it! Thanks for working with us on this @jgowdy!

PR #330, opened 2022-03-14
Upgrade pygit2 to allow version until <2.0

paulodiovani:
Current `tartufo` version requires dependency `pygit2` <1.7, which depends on `libgit2` `v1.1.x` (see https://github.com/libgit2/pygit2/blob/master/CHANGELOG.rst).
Most recent operating systems packages include `libgit2` v1.2 or newer (newest version is `1.4.2`), which would break the installation of `tartufo`.

This PR updates `pygit2` dependency to allow until version `<2.0`, which should be ok as no breaking changes have been added until `v1.9`.

--------

* [ ] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [x] No (breaking changes)

### Breaking change notice

- `pygit2` `v1.8.0` introduces a breaking change by renaming `RemoteCallbacks.progress()` to `RemoteCallbacks.sideband_progress()`, but I didn't find any ocurrence of this function call in the source code.

## What's new?
- Compatible to `pygit2` from versions `1.6` to `1.9.x`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Features:
 
 * [#328](https://github.com/godaddy/tartufo/pull/328) - Buffer issues beyond --buffer-size to a temporary file
 
+v3.0.0 - 15 March 2022
+----------------------
+
+Allow pygit2 dependency <=1.6,>2
+
 v3.0.0 - 5 January 2022
 -----------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,9 @@ Features:
 
 * [#328](https://github.com/godaddy/tartufo/pull/328) - Buffer issues beyond --buffer-size to a temporary file
 
-v3.0.0 - 15 March 2022
-----------------------
+Bug Fixes:
 
-Allow pygit2 dependency <=1.6,>2
-
+* [#330](https://github.com/godaddy/tartufo/pull/330) - Allow newer versions of pygit2 for newer versions of Python
 v3.0.0 - 5 January 2022
 -----------------------
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3-alpine3.14 as base
+FROM python:3-slim as base
 
 WORKDIR /app
 
@@ -9,9 +9,6 @@ ENV PIP_DEFAULT_TIMEOUT=100 \
     PIP_NO_CACHE_DIR=1 \
     POETRY_VERSION=1.1.12
 
-RUN apk update && \
-    apk upgrade && \
-    apk add --no-cache cargo gcc libffi-dev musl-dev openssl-dev rust libgit2-dev
 RUN pip --no-cache-dir install "poetry==$POETRY_VERSION"
 RUN python -m venv /venv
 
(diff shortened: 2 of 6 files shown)


tarkatronic:
Unfortunately this is not a possible upgrade at the moment. We had to make the decision at the time to pin `pygit2` to an older version, as we had to maintain Python 3.6 support, and this was the newest version of `pygit2` that allowed that.

It would be great to drop Python 3.6 support, but that will require a major version bump (v4.0) which requires a lot harder consideration.

One thing that has been discussed, but not yet experimented with, is that we might possibly be able to require different versions of `pygit2` depending on the Python version. If this works, it could be a very nice solution.

rbailey-godaddy:
I merged a commit that twiddles the dependency along the line @tarkatronic suggested, together with adding python 3.10 to the unit test matrices. My PC doesn't have fully-baked environments for all of these versions available, so we'll need to see what verdict is returned by the CI tests and then potentially do some followup triage.

paulodiovani:
> I merged a commit that twiddles the dependency along the line @tarkatronic suggested, together with adding python 3.10 to the unit test matrices. My PC doesn't have fully-baked environments for all of these versions available, so we'll need to see what verdict is returned by the CI tests and then potentially do some followup triage.

I was doing the multiple dependency change and when I pushed you havea lready done it :sweat_smile: 

Good job. And thanks.

rbailey-godaddy:
> Good job. And thanks.

Thank me when it works. I am about to commit a further change to rebase the docker image onto 3-slim (Debian) instead of alpine, as this significantly simplifies the build and eliminates weird libgit2 dependencies (because we can install wheel packages).

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ Features:
 
 * [#328](https://github.com/godaddy/tartufo/pull/328) - Buffer issues beyond --buffer-size to a temporary file
 
+v3.0.0 - 15 March 2022
+----------------------
+
+Allow pygit2 dependency <=1.6,>2
+
```
tarkatronic:
```suggestion
Bug Fixes:

* [#330](https://github.com/godaddy/tartufo/pull/330) - Allow newer versions of pygit2 for newer versions of Python
```
This is a bit more of an accurate/holistic description of the change, and also sticks to the surrounding formatting 😄

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3-alpine3.14 as base
+FROM python:3-slim as base
```
tarkatronic, line 1: What is the resulting change in image size here?

tarkatronic (changes requested):
Just one small change and one question here -- thanks for putting in this work, this is going to help a lot of people!

Dockerfile
rbailey-godaddy: `3-alpine3.14` is 45.2 MB; `3-slim` is 122 MB -- the final image is 260 MB, which is noticeably larger than the 88.1 MB of the current alpine-based build from `main`.

On the other hand, it's not quite apples-to-apples because the current code (in this branch) can't be built on any existing version of alpine at all, because compilation requires newer versions of some libraries than anything shipped by alpine. We could potentially fudge this by backleveling versions (to less backleveled than they were) in hopes of finding some happy medium, but it just didn't seem worth it.

That was before I read https://pythonspeed.com/articles/base-image-python-docker-images/ yesterday. :)

tarkatronic (approved):

smimani-godaddy (approved):
LGTM 🦈

sushantmimani (approved):

Issue #350, opened 2022-04-13
No feature parity in detection between pre-commit and scan-local-repo for tartufo v3, contrary to v2


pmevzek-godaddy:
## 🐛 Bug Report

My expectation is that pre-commit on a given commit and scan-local-repo just after the commit is added returns the same results.
If `pre-commit` says "all clear", and the commit is added, if I run `scan-local-repo` just after, it shouldn't flag as a problem the exact same commit that `pre-commit` was happy with.

However that does not seem the case, `pre-commit` lets a commit go through that is right after flagged by `scan-local-repo`.

## To Reproduce

```
$ git init test3
Initialized empty Git repository in /private/tmp/test3/.git/
$ cd test3
$ git commit --allow-empty -m 'Start'
[main (root-commit) 42f5966] Start
$ tartufo --version
tartufo, version 3.1.3
$ echo OnVybD48c21kOnZvaWNlPiszMi4yMDAwMDAwMDwvc21kOnZvaWNlPjwvc21kOmlzc3VlckluZm8 > secret.txt
$ git add secret.txt
$ tartufo --entropy pre-commit
Time: 2022-04-13T14:53:40.109393
All clear. No secrets detected.
$ git commit -m 'Secret added'
[main 15032bb] Secret added
 1 file changed, 1 insertion(+)
 create mode 100644 secret.txt
$ tartufo --entropy scan-local-repo .
~~~~~~~~~~~~~~~~~~~~~
Reason: High Entropy
Filepath: secret.txt
Signature: ca29177c396aa5465f41495af1e486d666308b51b7dab52228730624466cbc25
Commit time: 2022-04-13 14:55:08
Commit message: Secret added

Commit hash: 15032bb925d8650ebd15fe8d5f3a2f8b76327168
Branch: main
diff --git a/secret.txt b/secret.txt
new file mode 100644
index 0000000..049ae76
--- /dev/null
+++ b/secret.txt
@@ -0,0 +1 @@
+OnVybD48c21kOnZvaWNlPiszMi4yMDAwMDAwMDwvc21kOnZvaWNlPjwvc21kOmlzc3VlckluZm8
```

Switching tartufo back to some 2.x versions make it work correctly, so something changed between versions 2 and 3.

```
$ pipx inject tartufo 'tartufo<3'
  injected package tartufo into venv tartufo
done! ✨ 🌟 ✨
$ pipx inject tartufo 'click<8'
  injected package click into venv tartufo
[...]

PR #352, opened 2022-04-27
Fix pre-commit ignoring untracked files

### LINKED ISSUE #331: Pre-commit hook: High entropy strings not detected in new files by 3.0.0
## 🐛 Bug Report

Tartufo 3.0.0 does not detect high entropy strings in files that are getting added to the index by the commit that is being evaluated.

## To Reproduce

1. Add/update pre-commit config file to use tartufo 3.0.0
```yaml
repos:
  - repo: https://github.com/godaddy/tartufo
    rev: v3.0.0
    hooks:
      - id: tartufo
        name: tartufo
        entry: tartufo
        args: [ pre-commit ]
        language: python
        pass_filenames: false
```
2. Install the pre-commit configuration
```bash
pre-commit install
```
3. Create a new file with high entropy string
4. Stage the file and commit it. Commit suceeeds
```bash
date | shasum > new_file.txt
git add new_file.txt
git commit -m "committing a high entropy string"
```

## Expected Behavior

Commit should not succeed, and return failure. (2.10.0 rejects the commit as expected)

## Environment

1. Macintosh
2. Installed pre-commit through pip (2.17.0)

jwilbur-godaddy:

* [ ] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [x] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking

closes #331 

## What's new?

- Fix non-scanning of new files. It just took a few extra git flags to show untracked files in the diff against `HEAD`.

### DIFF
--- a/tests/test_scan_local_repo.py
+++ b/tests/test_scan_local_repo.py
@@ -8,6 +8,7 @@ from pygit2 import Repository
 from tartufo import cli, types
 from tests import helpers
 
+
 class ScanLocalRepoTests(unittest.TestCase):
     @mock.patch("tartufo.commands.scan_local_repo.GitRepoScanner")
     def test_scan_exits_gracefully_on_scan_exception(
@@ -43,13 +44,13 @@ class ScanLocalRepoTests(unittest.TestCase):
 
         # Check that tartufo picks up on newly added files
         repo.index.add("tests/data/config/" + file_name)
-        repo.index.write() # This actually writes the index to disk. Without it, the tracked file is not actually staged.
+        repo.index.write()  # This actually writes the index to disk. Without it, the tracked file is not actually staged.
         result = runner.invoke(cli.main, ["--entropy-sensitivity", "1", "pre-commit"])
         self.assertNotEqual(result.exit_code, 0)
 
         # Cleanup
         repo.index.remove("tests/data/config/" + file_name)
-        repo.index.write() # This actually writes the index to disk. Without it, secret.key will not removed from the index.
+        repo.index.write()  # This actually writes the index to disk. Without it, secret.key will not removed from the index.
         remove(file_name)
 
     def test_new_unstaged_file_does_not_show_up(self):
--- a/tests/test_scan_local_repo.py
+++ b/tests/test_scan_local_repo.py
@@ -1,13 +1,13 @@
 import unittest
 from pathlib import Path
 from unittest import mock
-
 from click.testing import CliRunner
-
 from tartufo import cli, types
-
 from tests import helpers
-
+from hashlib import sha256
+from pygit2 import Repository
+from subprocess import call
+from os import remove
 
 class ScanLocalRepoTests(unittest.TestCase):
     @mock.patch("tartufo.commands.scan_local_repo.GitRepoScanner")
@@ -32,3 +32,27 @@ class ScanLocalRepoTests(unittest.TestCase):
                 str(result.exception),
                 f"Repository not found at {Path(run_path).resolve()}",
             )
+
+    def test_new_file_shows_up(self):
+        runner = CliRunner()
+        # Add file with high entropy
+        secret_key = sha256(b"hello world")
+        f = open("secret.key", "a")
+        f.write(secret_key.hexdigest())
+        f.close()
+        repo = Repository(".")
+
+        # Check that tartufo picks up on newly added files
+        repo.index.add("tests/data/config/secret.key")
+        repo.index.write() # This actually writes the index to disk. Without it, the tracked file is not actually staged.
+        result_1 = runner.invoke(cli.main, ["--entropy-sensitivity", "1", "pre-commit"])
+        self.assertNotEqual(result_1.exit_code, 0)
+
+        # Check that tartufo does not flag non-added new files
+        repo.index.remove("tests/data/config/secret.key")
+        repo.index.write() # This actually writes the index to disk. Without it, secret.key will not return to being untracked.
+        result_2 = runner.invoke(cli.main, ["--entropy-sensitivity", "1", "pre-commit"])
+        self.assertEqual(result_2.exit_code, 0)
+
+        # Cleanup
+        remove("secret.key")
--- a/tests/test_scan_local_repo.py
+++ b/tests/test_scan_local_repo.py
@@ -1,13 +1,12 @@
 import unittest
 from pathlib import Path
 from unittest import mock
+from hashlib import sha256
+from os import remove
 from click.testing import CliRunner
+from pygit2 import Repository
 from tartufo import cli, types
 from tests import helpers
-from hashlib import sha256
-from pygit2 import Repository
-from subprocess import call
-from os import remove
 
 class ScanLocalRepoTests(unittest.TestCase):
     @mock.patch("tartufo.commands.scan_local_repo.GitRepoScanner")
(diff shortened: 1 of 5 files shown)


rbailey-godaddy:
This commit is admirably simple but hurts my brain to think about it. A few questions/requests/comments...
* Can you add a comment citing https://github.com/libgit2/libgit2/blob/main/include/git2/diff.h so people who want to understand these flags can quickly find information about what they do?
* I believe I understand why this works, but I don't understand why it is necessary. Both #331 and #350 provide test cases that explicitly do a `git add` before running the pre-commit check; at this point, the bad file should be in the index and it seems the existing `cached=True` should be sufficient to do what we want, i.e. "the index is compared to 'a' ["HEAD"]". Additionally tossing untracked files into the mix seems problematic logically in the sense that we do not want to fail a pre-commit check on the basis of files the caller has not added to the commit.

This is perhaps a bit expansive for a unit test, but can you validate the following scenario?

```sh
date | shasum > file-a.txt
date | shasum > file-b.txt
git add file-a.txt
tartufo pre-commit
```
We expect that a finding will be generated for file-a.txt, but not for file-b.txt. I worry that the change as proposed at this point will flag both files.

jwilbur-godaddy:
> This commit is admirably simple but hurts my brain to think about it. A few questions/requests/comments...
> 
> * Can you add a comment citing https://github.com/libgit2/libgit2/blob/main/include/git2/diff.h so people who want to understand these flags can quickly find information about what they do?
> * I believe I understand why this works, but I don't understand why it is necessary. Both [Pre-commit hook: High entropy strings not detected in new files by 3.0.0 #331](https://github.com/godaddy/tartufo/issues/331) and [No feature parity in detection between pre-commit and scan-local-repo for tartufo v3, contrary to v2 #350](https://github.com/godaddy/tartufo/issues/350) provide test cases that explicitly do a `git add` before running the pre-commit check; at this point, the bad file should be in the index and it seems the existing `cached=True` should be sufficient to do what we want, i.e. "the index is compared to 'a' ["HEAD"]". Additionally tossing untracked files into the mix seems problematic logically in the sense that we do not want to fail a pre-commit check on the basis of files the caller has not added to the commit.
> 
> This is perhaps a bit expansive for a unit test, but can you validate the following scenario?
> 
> ```shell
> date | shasum > file-a.txt
> date | shasum > file-b.txt
> git add file-a.txt
> tartufo pre-commit
> ```
> 
> We expect that a finding will be generated for file-a.txt, but not for file-b.txt. I worry that the change as proposed at this point will flag both files.

Hey Scott,

I will add that comment you requested. That's a good idea. As for the rest, I have a test case that I have shared elsewhere, which I will paste below. For clarification, this does not scan all files in the repo. Because of the recent addition of `cached=True` on the same exact line that I changed, this will only apply to untracked files that were added to the index, meaning that this will _not_ result in tartufo scanning unstaged changes. I think the problem is that a file still counts as "untracked" even after it has been added to the index. It only becomes tracked once it has actually been committed.

### Test Case

1. Clone tartufo
2. Run openssl genrsa > private_key.pem to place an RSA private key right in the tartufo repository. (If you do this in Powershell, the file will be UTF-16-LE encoded, so tartufo will NOT catch this (I am going to report this as a bug, too). You can re-encode it in VS Code.)
3. After ~line 571 (for chunk in self.chunks) in scanner.py, add print(chunk.file_path) so you can see what files are being scanned.
4. One line ~909 of scanner.py, remove the flags.
5. Run poetry run python .\tartufo\_main_.py pre-commit. Observe that it DOES NOT scan private_key.pem. This is because there are no flags telling it to include untracked files and their contents.
6. One line ~909 of scanner.py, reintroduce the flags.
7. Run the command from step 5 again. Observe that it DOES NOT scan private_key.pem. This is because private_key.pem is not in the Git index.
8. Stage private_key.pem.
9. Run the command from step 5 again. Observe that it DOES scan private_key.pem because it is now staged.
"does that cause us to screen non-staged untracked files as well"
No, it does not, because of the cached=True flag that was introduced recently on that same line.

rbailey-godaddy:
The lightbulb goes on...

>  I think the problem is that a file still counts as "untracked" even after it has been added to the index. It only becomes tracked once it has actually been committed.

This was totally non-obvious to me. Can you extend your comment to include this bit by way of explanation regarding why the chosen flags are present? Otherwise, I'm happy.

rbailey-godaddy (approved):
But I'm expecting to see that those URLs will trigger entropy findings that will need to be added to pyproject.toml lol...

rbailey-godaddy:
@jwilbur-godaddy can you cherry-pick 78b27ee0699114d1b865070f0d47c5eb0d806b24 from the tartufo repo into your branch? This should fix the issues I noted earlier in my review.

pmevzek-godaddy:
> This commit is admirably simple but hurts my brain to think about it.

I fully agree with this, and personally I wouldn't see it favorably this gets merged (even if it seems to solve a problem, I still think there is a code-smell here).

At the very least, there is a terminology problem. A new file being staged/cached IS NOT UNTRACKED.
This comes straight from `git-status` manual:

>    Short Format
>        In the short-format, the status of each path is shown as one of these forms
> 
>            XY PATH
>            XY ORIG_PATH -> PATH
> 
> ...
> 
>            X          Y     Meaning
>            -------------------------------------------------
>                     [AMD]   not updated
>            M        [ MTD]  updated in index
>            T        [ MTD]  type changed in index
>            A        [ MTD]  added to index
>            D                deleted from index
>            R        [ MTD]  renamed in index
>            C        [ MTD]  copied in index
>            [MTARC]          index and work tree matches
>            [ MTARC]    M    work tree changed since index
>            [ MTARC]    T    type changed in work tree since index
>            [ MTARC]    D    deleted in work tree
>                        R    renamed in work tree
>                        C    copied in work tree
>            -------------------------------------------------
>            D           D    unmerged, both deleted
>            A           U    unmerged, added by us
>            U           D    unmerged, deleted by them
>            U           A    unmerged, added by them
>            D           U    unmerged, deleted by us
>            A           A    unmerged, both added
>            U           U    unmerged, both modified
>            -------------------------------------------------
>            ?           ?    untracked
>            !           !    ignored
>            -------------------------------------------------
> 

Now let us try with a new file being committed:

```
$ git init test
$ cd test
$ git status --porcelain
$

# empty, as expected

$ > a_new_file
$ git status --porcelain
?? a_new_file

# per manual above, THIS is an untracked file for now

# let us stage it now

$ git add a_new_file
$ git status --porcelain
A  a_new_file

# per documentation above, the file is "ADDED" it is NOT untracked anymore.
```

Maybe the confusion is coming from `git` and/or `libgit2` itself (maybe the flag has to be read as "untracked OR newly added" and not just "untracked"), but at the very least I don't think it is right to say that a new added file (to cache/stage) is untracked. It is not untracked anymore once staged.

jwilbur-godaddy:
Okay, so do you just want me to change the comment? What are you asking for?

mdayanc-GoDaddy (approved):

hong-godaddy:
Can we get additional reviews on this? Would love to get this merged and have a release

jgowdy:
Do we have a unit test we can run against the existing main branch that fails to detect something in the index that is untracked, and that with these new flags it passes, and do we have a unit test that demonstrates that with the flags that untracked files not in the index are NOT scanned?  I see the steps to reproduce above, but we should codify those in tests to validate the change we are making.

jgowdy:
This also needs a changelog entry per the PR template

jgowdy:
Can we test the negative case also?  i.e. That this change doesn't cause untracked files not staged to the index to be scanned?

tarkatronic:
> Can we test the negative case also? i.e. That this change doesn't cause untracked files not staged to the index to be scanned?

I've got to agree with Jeremiah on this one. I really like this solution, and its elegance, but I definitely think we want that other test just to be extra careful and cautious. Looking at the test you already wrote, it looks like it should even be a decently simple one to add!

jwilbur-godaddy:
@jgowdy @tarkatronic I don't understand how this section of code does not deliver what you are asking for:

https://github.com/jwilbur-godaddy/tartufo/blob/bcc2716fe10757808fa48e097bb1ab911a898b8c/tests/test_scan_local_repo.py#L51-L55

tarkatronic:
> 

Ah ha. I had missed that part. Can you split that out to a separate unit test so that the two conditions are tested (and can potentially fail) on their own?

tarkatronic:
One other thing, it looks like `black` and `pylint` are failing in the CI pipeline -- could you take a look and see if you can get those passing locally? Then we can get this merged and released!

tests/test_scan_local_repo.py
```diff
@@ -32,3 +31,36 @@ def test_scan_exits_gracefully_when_target_is_not_git_repo(self):
...
+        secret_key = sha256(b"hello world")
+        with open(file_name, "a") as file:
+            file.write(secret_key.hexdigest())
+        repo = Repository(".")
+
+        # Check that tartufo picks up on newly added files
+        repo.index.add("tests/data/config/" + file_name)
+        repo.index.write() # This actually writes the index to disk. Without it, the tracked file is not actually staged.
+        result = runner.invoke(cli.main, ["--entropy-sensitivity", "1", "pre-commit"])
+        self.assertNotEqual(result.exit_code, 0)
```
tarkatronic, line 49: I finally realized what struck me as odd here. An exit code of `0` would indicate a successful scan with no issues found. So what I'm not sure about here is, how does this verify that the file was scanned? Since you're writing a sha256 digest to the file, that should get flagged for entropy, though I notice you modify the entropy sensitivity. Shouldn't we be letting it fail, and checking for an exit code greater than 0? Or even better maybe capture json output from the command and ensure that the filename in question is in the output?

jwilbur-godaddy: Did you miss the `assertNotEqual`? I am checking that the return code is _not_ zero.

tarkatronic: Oops haha yes I did, sorry. ☕ has clearly not kicked in yet!

tarkatronic (approved):
Looks great to me! Thanks for all the work and putting up with us through the process @jwilbur-godaddy!

@jgowdy can you take another look and make sure your concerns are covered?

jgowdy (approved):
Black still doesn't seem happy, but my concerns about tests are satisfied.  Can we fix the black linting and then we're g2g

jwilbur-godaddy:
Done! Thank you, all!

rbailey-godaddy (approved):

jwilbur-godaddy:
Just a heads up, I am starting paternity leave tomorrow, so I will not be able to follow up on this after today. I also do not have write access to this repo, so I will not be able to merge this in.

Issue #347, opened 2022-04-07
matched_string contains "+" character, because of that BFG is unable to clean git history


sdzialowski-godaddy:
## 🐛 Bug Report

Report generated by tartufo in the field matched_string is preceded with "+" character which is part of git diff and later causes problems when those secret are passed through [BFG](https://rtyley.github.io/bfg-repo-cleaner/) tool

## To Reproduce

Let's create empty directory (somewhere, up to you where) and let's enter it:

    mkdir xxx
    cd xxx
    
    
Make sure you have installed [jq](https://stedolan.github.io/jq/) (on mac)

    HOMEBREW_NO_AUTO_UPDATE=1 brew install jq
    
Then let's create script which will serve as a tool to crate local git repository with particular history.
(Just copy all below code and paste it into your terminal as is)

```bash

cat <<EOF > create_bare_repository.sh

set -e
set -x

rm -rf repository
rm -rf clone
rm -rf mirror

# let's create new clean local bare repository
mkdir repository
(
  cd repository
  git init --bare
)

# then clone it to separeate directory
git clone ./repository clone
(
  cd clone

  # then let's add a file with secret
  cat <<EEE > file_with_secret.txt
Some high entroply data:
end of file
EEE

  git add .
  git commit -m "first commit"

  cat <<EEE > file_with_secret.txt
Some high entroply data:
igv0nCsCQQCKF8jALudfgq1bRCkB8Cq6fgCEon9N8CdV1pT9ImqR7LAi6aVbjPnq
lI9rMdEPUtWn5RLT8zpxQyhdDFiMYYCjAkBMCyHeQppTmhkgfFu64rcRqCGskSER
6QxrIYbOdhyXggFLayQjK9FcX7n6GhrKRWY3lBfUpJAbiTzufEfi0PDdAkA5qsUC
end of file
EEE

  git add .
  git commit -m "second commit"

  rm -rf file_with_secret.txt

  git add .
  git commit -m "third commit"

  # push everything
  git push origin master

  git log
)


EOF

```

then let's use it to create bare repository with some regular secret

    /bin/bash create_bare_repository.sh
    
then let's make the mirror of that repository as it is suggested in [BFG documentation](https://rtyley.github.io/bfg-repo-cleaner/#usage)

    git clone --mirror ./repository ./mirror

Then let's run tartufo through docker against that ./mirror directory

    docker run --rm -v "$(pwd)/mirror:/git" godaddy/tartufo --output-dir /git/.tartufo \
[...]

PR #368, opened 2022-07-17
Add update-signatures command to migrate deprecated signatures

jhall1-godaddy:

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [x] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?

This pull request addresses the recent deprecation of some signatures which were generated with the leading `+`/`-` from the git diff. That issue is documented [here](https://github.com/godaddy/tartufo/issues/347).

As stated by the deprecation warning, in tartufo v4.X those signatures will no longer be valid.
In an attempt to ease the migration for all parties involved, this command allows for migration with a single command.

The command performs a local scan (we assume the user is scanning locally, this needs to be documented), parses the deprecations from the ouput there, and makes adjustments in the end users tartufo config file.

Help snippet:
```
Commands:
  scan-local-repo    Scan a repository already cloned to your local system.
  scan-folder        Scan a folder.
  update-signatures  Update deprecated signatures for a local repository.
  pre-commit         Scan staged changes in a pre-commit hook.
  scan-remote-repo   Automatically clone and scan a remote git repository.
```

Invoked identically to the `scan-local-repo` command, it accepts all the same options as well.
- E.G. `tartufo update-signatures .`

---

### Usage

I used tartufo itself for testing, and will include my results here. 
Prior to updating the signatures tartufo was producing the following deprecations:

```
/home/jonx/projects/work/tartufo/tartufo/scanner.py:661: DeprecationWarning: Signature 358a1ae040ab71b3aafeb4318e5d9e318ff1b681c5bf0e99d381584a33ee7833 was generated by an old version of tartufo and is deprecated. tartufo 4.x will not recognize this signature. Please update your configuration to use signature 38d3398905c382a79857e884f33062a5a38e4e3f8b5d54f269879601b2e847fc instead.
  warnings.warn(
/home/jonx/projects/work/tartufo/tartufo/scanner.py:661: DeprecationWarning: Signature 358a1ae040ab71b3aafeb4318e5d9e318ff1b681c5bf0e99d381584a33ee7833 was generated by an old version of tartufo and is deprecated. tartufo 4.x will not recognize this signature. Please update your configuration to use signature 38d3398905c382a79857e884f33062a5a38e4e3f8b5d54f269879601b2e847fc instead.
  warnings.warn(
/home/jonx/projects/work/tartufo/tartufo/scanner.py:661: DeprecationWarning: Signature 6c70b4a712ada69f2a6c919edc783cf485b6637e3c5429de376ef1a451a042e4 was generated by an old version of tartufo and is deprecated. tartufo 4.x will not recognize this signature. Please update your configuration to use signature d26b223efb2087d4a3db7b3ff0f65362372c46190fd74b8061db8099f4a2ee60 instead.
  warnings.warn(
/home/jonx/projects/work/tartufo/tartufo/scanner.py:661: DeprecationWarning: Signature 357841766ca69979608e4e7fb1da2addeafe6d15cade15466c39d903cc0488a8 was generated by an old version of tartufo and is deprecated. tartufo 4.x will not recognize this signature. Please update your configuration to use signature 94c45f9acae0551a2438fa8932fac466e903fd9a09bd2e11897198b133937ccd instead.
  warnings.warn(
/home/jonx/projects/work/tartufo/tartufo/scanner.py:661: DeprecationWarning: Signature 88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5 was generated by an old version of tartufo and is deprecated. tartufo 4.x will not recognize this signature. Please update your configuration to use signature 482e7b31188f1e59dd6eb5b21c46f7467449ffcbeba9a4ca412dfcd8ea9ef66f instead.
  warnings.warn(
/home/jonx/projects/work/tartufo/tartufo/scanner.py:661: DeprecationWarning: Signature 23996f9309c042d37958d28c2cc0c16a060a03376ee075b5b077609e2a299e44 was generated by an old version of tartufo and is deprecated. tartufo 4.x will not recognize this signature. Please update your configuration to use signature 482e7b31188f1e59dd6eb5b21c46f7467449ffcbeba9a4ca412dfcd8ea9ef66f instead.
  warnings.warn(
/home/jonx/projects/work/tartufo/tartufo/scanner.py:661: DeprecationWarning: Signature 85109edb8081a60f3a44139a98b643b8cacb777f4d5c814276ae729d50fe414c was generated by an old version of tartufo and is deprecated. tartufo 4.x will not recognize this signature. Please update your configuration to use signature bf48e316ecad0a37071ed81cc54a6b629a1a3bf73cf52d06a77c9fbaf91bc833 instead.
  warnings.warn(
Time: 2022-07-17T13:36:54.860205
```

---

Running the command:

```bash
$ tartufo update-signatures .                                                                                                                                                                            
Found 6 unique deprecated signatures.
1) Updating '23996f9309c042d37958d28c2cc0c16a060a03376ee075b5b077609e2a299e44' -> '482e7b31188f1e59dd6eb5b21c46f7467449ffcbeba9a4ca412dfcd8ea9ef66f'
2) Updating '6c70b4a712ada69f2a6c919edc783cf485b6637e3c5429de376ef1a451a042e4' -> 'd26b223efb2087d4a3db7b3ff0f65362372c46190fd74b8061db8099f4a2ee60'
3) Updating '88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5' -> '482e7b31188f1e59dd6eb5b21c46f7467449ffcbeba9a4ca412dfcd8ea9ef66f'
4) Updating '358a1ae040ab71b3aafeb4318e5d9e318ff1b681c5bf0e99d381584a33ee7833' -> '38d3398905c382a79857e884f33062a5a38e4e3f8b5d54f269879601b2e847fc'
5) Updating '357841766ca69979608e4e7fb1da2addeafe6d15cade15466c39d903cc0488a8' -> '94c45f9acae0551a2438fa8932fac466e903fd9a09bd2e11897198b133937ccd'
6) Updating '85109edb8081a60f3a44139a98b643b8cacb777f4d5c814276ae729d50fe414c' -> 'bf48e316ecad0a37071ed81cc54a6b629a1a3bf73cf52d06a77c9fbaf91bc833'
Updated 6 total deprecated signatures.
```

---

After running the command, the pyproject.toml was altered as follows:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 8088fce..0e50d97 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,8 +96,8 @@ exclude-signatures = [
   {signature = "dbc084c22395931afd734a5be53f1cde913c7d988034b856087c744b424d9673", reason = "High entropy string BASE64_CHARS in truffleHog/truffleHog.py 9-Jan-2017"},
   {signature = "bc67523b08b2170ca9b802535684454140679af3156bff808508cf32b33aa240", reason = "High entropy string random_stringHex in tests.py 28-Sep-21017"},
   {signature = "978dc8605ef9bf471e467dd2d8570fd658f0a5f25378799a2ad8bb6ea480173d", reason = "High entropy string random_stringB64 in tests.py 28-Sep-21017"},
-  {signature = "85109edb8081a60f3a44139a98b643b8cacb777f4d5c814276ae729d50fe414c", reason = "High entropy string in secretFile.txt 30-Dec-2016"},
-  {signature = "23996f9309c042d37958d28c2cc0c16a060a03376ee075b5b077609e2a299e44", reason = "High entropy string in temp/nothing 30-Dec-2016"},
+  {signature = "bf48e316ecad0a37071ed81cc54a6b629a1a3bf73cf52d06a77c9fbaf91bc833", reason = "High entropy string in secretFile.txt 30-Dec-2016"},
+  {signature = "482e7b31188f1e59dd6eb5b21c46f7467449ffcbeba9a4ca412dfcd8ea9ef66f", reason = "High entropy string in temp/nothing 30-Dec-2016"},
   {signature = "51dad1be0950e627bb3660a6246b42ddc6d6d431267b309af2469114fd2df2e5", reason = "Regular expression match, RSA private key in truffleHog/regexChecks.py 8-Dec-2017"},
   {signature = "74c056ae650ffb202ea3b68207dce82ebeae0414cfe3b49f9db0c79841473da1", reason = "Regular expression match, PGP private key block in truffleHog/regexChecks.py 8-Dec-2017"},
   {signature = "36617c71d89b78193c99fdd2a28027319dfdb361d649372bc72c45317f9edf24", reason = "Regular expression match, SSH (EC) private key in truffleHog/regexChecks.py 8-Dec-2017"},
@@ -131,13 +131,13 @@ exclude-signatures = [
   {signature = "586f9df1870fcfafd3070dc553e97db25b3fce82c9144602567d69a86597c06f", reason = "High entropy string commit_w_secret in tests/test_scanner.py 11-Nov-2019"},
   {signature = "842533f44cf32fb3d93a7f56227977aa4f16caafe82ace8f5f4de27d750c1ec1", reason = "High entropy string since_commit in tests/test_scanner.py 11-Nov-2019"},
   {signature = "d039c652f27c4d42026c5b5c9be31bfe368b283b24248e98d92f131272580053", reason = "High entropy string BASE64_CHARS in tartufo/scanner.py 25-Aug-2020"},
-  {signature = "6c70b4a712ada69f2a6c919edc783cf485b6637e3c5429de376ef1a451a042e4", reason = "tests/data/scan_folder/donotscan.txt"},
-  {signature = "357841766ca69979608e4e7fb1da2addeafe6d15cade15466c39d903cc0488a8", reason = "tests/data/scan_folder/test.txt"},
+  {signature = "d26b223efb2087d4a3db7b3ff0f65362372c46190fd74b8061db8099f4a2ee60", reason = "tests/data/scan_folder/donotscan.txt"},
+  {signature = "94c45f9acae0551a2438fa8932fac466e903fd9a09bd2e11897198b133937ccd", reason = "tests/data/scan_folder/test.txt"},
   {signature = "743b893e30777d9c4e28d3d341ca4ba1c2541a9c62b497dece75c2a64420e770", reason = "tests/test_folder_scanner.py"},
-  {signature = "358a1ae040ab71b3aafeb4318e5d9e318ff1b681c5bf0e99d381584a33ee7833", reason = "tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt"},
+  {signature = "38d3398905c382a79857e884f33062a5a38e4e3f8b5d54f269879601b2e847fc", reason = "tests/data/scan_folder/scan_sub_folder/sub_folder_test.txt"},
   {signature = "9d1b278fa384c9dc58d607130d97740910309d13b29f13f7ed6348738ea4f32c", reason = "tests/test_folder_scanner.py 16-Nov-2021"},
   {signature = "9f675d4d7f43e484e07fb199a1e7ec372ef204631c1f4ec054acc3119fbde4cf", reason = "build_and_publish_docker_file 13-Jun-2019"},
-  {signature = "88152f1d10077f417ee5aea5fa86adaa5b74d3f2de2e87ce4b4aa5cd79c7b2f5", reason = "nothing file, 30-Dec-2016"},
+  {signature = "482e7b31188f1e59dd6eb5b21c46f7467449ffcbeba9a4ca412dfcd8ea9ef66f", reason = "nothing file, 30-Dec-2016"},
   {signature = "c94ba36a7411bf89fab6e87076740deecdc7a61fcfbbc1aa5934608f6c4f3adb", reason = "tests/data/config/rule_pattern_config.toml"},
   {signature = "4f44b817b42c94fa01f47166ca7adaa0b750b1b6cd84f7ce3bf23a7728cdac57", reason = "commit hashes in github URLs in comments, tartufo/scanner.py"}
 ]
 ```

---

### Thoughts

- Tests should be feature complete, with full coverage.
- I took a more functional approach here because frankly this command didn't feel like it would benefit from OOP in any way. 
- I haven't explicitly written any documentation other than the stuff added by click and sphinx automatically from docstrings, I am interested in getting feedback on how that should be structured.

### DIFF
--- a/tartufo/commands/update_signatures.py
+++ b/tartufo/commands/update_signatures.py
@@ -29,7 +29,7 @@ def scan_local_repo(
     max_depth: int,
     branch: Optional[str],
     include_submodules: bool,
-) -> Tuple[Optional[GitRepoScanner], io.StringIO, io.StringIO]:
+) -> Tuple[Optional[GitRepoScanner], io.StringIO]:
     """A reworked version of the scan-local-repo command.
 
     :param options: The options provided to the top-level tartufo command
@@ -62,7 +62,8 @@ def scan_local_repo(
             except types.TartufoException as exc:
                 click.echo(util.style_error(str(exc)), err=True)
 
-    return scanner, stdout, stderr
+    del stdout
+    return scanner, stderr
 
 
 def get_deprecations(stderr: io.StringIO) -> DeprecationSetT:
@@ -105,7 +106,7 @@ def replace_deprecated_signatures(
         # they are found in the exclude-signatures section of config
         for target_signature in filter(targets, config_data["exclude_signatures"]):
             updated += 1
-            click.echo(f"{updated}) Updating {old_sig!r} -> {new_sig!r}")
+            click.echo(f"{updated}) {old_sig!r} -> {new_sig!r}")
             target_signature["signature"] = new_sig
 
     return updated
@@ -131,6 +132,27 @@ def write_updated_signatures(
         file.write(tomlkit.dumps(result))
 
 
+def remove_duplicated_entries(config_data: MutableMapping[str, Any]) -> int:
+    """Removed any duplicated signature entries in the config.
+    Keeps the first instance of each duplicate.
+
+    :param config_data: The config data to check for duplicates
+    """
+    seen = set()
+    count = 0
+
+    for i, exclude in enumerate(config_data["exclude_signatures"].copy()):
+        if exclude["signature"] in seen:
+            # Remove this duplicated signature
+            del config_data["exclude_signatures"][i]
+            count += 1
+        else:
+            # Mark this signature as seen
+            seen.add(exclude["signature"])
+
+    return count
+
+
 @click.command("update-signatures")
 @click.option("--since-commit", help="Only scan from a given commit hash.", hidden=True)
 @click.option(
@@ -152,6 +174,20 @@ def write_updated_signatures(
     show_default=True,
     help="Controls whether the contents of git submodules are scanned",
 )
+@click.option(
+    "--update-configuration/--no-update-configuration",
+    is_flag=True,
+    default=True,
+    show_default=True,
+    help="Whether or not to overwrite the tartufo config file.",
+)
+@click.option(
+    "--remove-duplicates/--no-remove-duplicates",
+    is_flag=True,
+    default=True,
+    show_default=True,
+    help="Whether or not to remove duplicated signatures.",
+)
 @click.pass_obj
 @click.pass_context
 def main(
@@ -162,6 +198,8 @@ def main(
     max_depth: int,
     branch: Optional[str],
     include_submodules: bool,
+    update_configuration: bool,
+    remove_duplicates: bool,
 ) -> GitRepoScanner:
     """Update deprecated signatures for a local repository."""
     config_path, config_data = load_config_from_path(pathlib.Path(repo_path))
@@ -172,12 +210,9 @@ def main(
             code=0,
         )
 
-    scanner, stdout, stderr = scan_local_repo(
+    scanner, stderr = scan_local_repo(
         options, repo_path, since_commit, max_depth, branch, include_submodules
     )
-
-    del stdout  # We are discarding stdout from the scan-local-repo command
-    # Should we print it to the user instead?
 
     if not scanner:
         # Explicitly fail if we didn't get a scanner back
@@ -187,10 +222,16 @@ def main(
     click.echo(f"Found {len(deprecations)} unique deprecated signatures.")
     updated = replace_deprecated_signatures(deprecations, config_data)
 
-    if deprecations:
-        write_updated_signatures(config_path, config_data)
-        click.echo(f"Updated {updated} total deprecated signatures.")
+    dups = 0
+    if remove_duplicates:
+        dups = remove_duplicated_entries(config_data)
 
-    # We would have failed earlier so this assert is safe
-    assert scanner is not None
+    if update_configuration and (deprecations or dups):
+        # Only rewrite the config if we have altered it
+        write_updated_signatures(config_path, config_data)
+        plural = lambda n: "" if n == 1 else "s"
+
+        click.echo(f"Removed {dups} duplicated signature{plural(dups)}.")
+        click.echo(f"Updated {updated} total deprecated signature{plural(updated)}.")
+
     return scanner
--- a/tartufo/commands/update_signatures.py
+++ b/tartufo/commands/update_signatures.py
@@ -38,7 +38,7 @@ def scan_local_repo(
     :param max_depth: A maximum depth, or maximum number of commits back in history, to scan
     :param branch: A specific branch to scan
     :param include_submodules: Whether to also scan submodules of the repository
-    :returns: The number of items replaced in config_data
+    :returns: A tuple containing optional scanner, stdout, and stderr from the scan
     """
     stdout = io.StringIO()
     stderr = io.StringIO()
--- a/tartufo/commands/update_signatures.py
+++ b/tartufo/commands/update_signatures.py
@@ -84,6 +84,7 @@ def main(
             util.style_warning("No signatures found in configuration, exiting..."), ctx
         )
 
+    updated = 0
     stdout = io.StringIO()
     stderr = io.StringIO()
     deprecations = set()
@@ -115,12 +116,13 @@ def main(
             deprecations.add(match.groups())
 
     click.echo(f"Found {len(deprecations)} unique deprecated signatures.")
-    for (i, (old_sig, new_sig)) in enumerate(deprecations):
+    for old_sig, new_sig in deprecations:
         targets = functools.partial(lambda o, s: o == s["signature"], old_sig)
         # Iterate all the deprecations and update them everywhere
         # they are found in the exclude-signatures section of config
         for target_signature in filter(targets, config_data["exclude_signatures"]):
-            click.echo(f"{i + 1}) Updating {old_sig!r} -> {new_sig!r}")
+            updated += 1
+            click.echo(f"{updated}) Updating {old_sig!r} -> {new_sig!r}")
             target_signature["signature"] = new_sig
 
     # Read the current config, for clean overwrite
@@ -134,6 +136,9 @@ def main(
 
     with open(str(config_path), "w") as file:
         file.write(tomlkit.dumps(result))
+
+    if deprecations:
+        click.echo(f"Updated {updated} total deprecated signatures.")
 
     # We would have failed earlier so this assert is safe
     assert scanner is not None
--- a/tartufo/commands/update_signatures.py
+++ b/tartufo/commands/update_signatures.py
@@ -12,18 +12,31 @@ from tartufo import types, util
 from tartufo.config import load_config_from_path
 from tartufo.scanner import GitRepoScanner
 
+DeprecationSetT = t.MutableSet[t.Sequence[str]]
 
-def _scan_local_repo(
+
+def scan_local_repo(
     options: types.GlobalOptions,
     repo_path: str,
     since_commit: t.Optional[str],
     max_depth: int,
     branch: t.Optional[str],
     include_submodules: bool,
-) -> t.Union[GitRepoScanner, None]:
-    """We had to duplicate the scan local repo command callback, to
-    alter the exception logic a bit and prevent swallowing errors.
+) -> t.Tuple[t.Optional[GitRepoScanner], io.StringIO, io.StringIO]:
+    """A reworked version of the scan-local-repo command.
+
+    :param options: The options provided to the top-level tartufo command
+    :param repo_path: The local filesystem path pointing to the repository
+    :param since_commit: A commit hash to treat as a starting point in history for the scan
+    :param max_depth: A maximum depth, or maximum number of commits back in history, to scan
+    :param branch: A specific branch to scan
+    :param include_submodules: Whether to also scan submodules of the repository
+    :returns: The number of items replaced in config_data
     """
+    stdout = io.StringIO()
+    stderr = io.StringIO()
+    scanner = None
+
     git_options = types.GitOptions(
         since_commit=since_commit,
         max_depth=max_depth,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: tartufo/commands/update_signatures.py [hunk 1]
```diff
@@ -0,0 +1,196 @@
...
+            ctx,
+            code=0,
+        )
+
+    scanner, stdout, stderr = scan_local_repo(
+        options, repo_path, since_commit, max_depth, branch, include_submodules
+    )
+
+    del stdout  # We are discarding stdout from the scan-local-repo command
+    # Should we print it to the user instead?
```
rbailey-godaddy: We should not. :) I believe the current behavior is on-point for the purpose of remediating configuration files. In passing, it occurs to me that in the normal case where the repo is fine (has no issues detected) and only suffers from deprecated signatures, there should be very little output to discard in any event.

tartufo/commands/update_signatures.py
```diff
@@ -0,0 +1,196 @@
...
+
+    if not scanner:
```
rbailey-godaddy, line 217:
```suggestion
    if scanner is None:
```
I believe this is explicit enough to placate mypy without requiring the additional assertion below.

tartufo/commands/update_signatures.py
```diff
@@ -0,0 +1,196 @@
...
+    deprecations = get_deprecations(stderr)
+    click.echo(f"Found {len(deprecations)} unique deprecated signatures.")
+    updated = replace_deprecated_signatures(deprecations, config_data)
+
+    if deprecations:
+        write_updated_signatures(config_path, config_data)
+        click.echo(f"Updated {updated} total deprecated signatures.")
+
+    # We would have failed earlier so this assert is safe
+    assert scanner is not None
```
rbailey-godaddy:
```suggestion
```
Not needed, I believe. I can't think of a case where `bool(scanner)` is `False` except when `scanner` is `None`, which I propose you test for above.

rbailey-godaddy (commented):
This looks good to me, but I have an additional challenge based on your sample output. :)

The old algorithm could actually generate three different signatures for the exact same `secret` under these circumstances, when it appeared at the beginning of a line:

* When introducing/modifying the secret (where it could have been prefixed with `+` in error)
* When removing/modifying the secret (where it could have been prefixed with `-` in error)
* When the secret was adjacent to an unrelated change (and it could appear as part of the diff context, where it would be prefixed with a space)

Note in the last case, the space would not be considered part of the secret (because it is not in the base64 encoding alphabet) and the "old" signature actually is identical to the "new" signature.

Given this behavior, it's probably common that multiple old signatures may collapse to the same new signature (as demonstrated in your example). While this is totally acceptable, I can't help wondering if it is possible to detect the duplication and simply delete the original signature instead of replacing it with a duplicate new signature.

Regardless of technical merit, I also could see a case made for leaving the duplicates anyway in order to reduce confusion when modifying non-trivial configurations.

rbailey-godaddy:
One additional thought on this very nice piece of work. Could you attach a flag to `update-signatures` such as `--replace-deprecated-signatures` which triggers this behavior? (I am not in love with this option name and if you think of something you like better, use what you like.) I am trying to make room for additional future behavior such as `--remove-unused` or `--replace-deprecated-format` that would fit naturally into this verb.

For convenience, we could have `--replace-deprecated-signatures` be the default and let people specify `--no-replace-deprecated-signatures` when they want that.

Thanks again for a very timely contribution!

rbailey-godaddy:
Maybe `update-configuration` would be better than `update-signatures` with respect to my last comment?

tartufo/commands/update_signatures.py
jhall1-godaddy: I was thinking the same.

jhall1-godaddy: I'll clean up that comment.

jhall1-godaddy:
> I can't help wondering if it is possible to detect the duplication and simply delete the original signature instead of replacing it with a duplicate new signature.

This is an interesting thought. At the very least we could warn of any duplicates and let the user handle it. Removing them would be pretty trivial - but which one is the correct one to remove? Maybe best to let them decide.

jhall1-godaddy:
`update-configuration` feels fitting. Will get that done.

tartufo/commands/update_signatures.py
jhall1-godaddy: I have a suspicion the assert is an artifact of some time before the `if not scanner:` check was there and may not even be necessary.
I will play around with it. Your solution should definitely work in any case.

sushantmimani (commented):
I am happy with these changes once @rbailey-godaddy's comments have been addressed. In terms of documentation, I feel that addition of a new paragraph in `upgrading3.rst` explaining the new features and how to use it should be enough.

tartufo/commands/update_signatures.py
jhall1-godaddy: Oddly enough, it's not. Both `if not scanner:` and `if scanner is None:` *should* work, but they don't - mypy and pyright both complain.

Is there any reason were actually returning the scanner object from all the commands? Is there any benefit to that?

jhall1-godaddy: It appears to be because the type checkers cannot infer that the function doesn't continue after the `util.fail` function. 

I can do something like this and it will prevent any assert from ever running and also placates the type checkers:

```py
if not scanner:
    # Explicitly fail if we didn't get a scanner back
    util.fail(util.style_error("Unable to update signatures"), ctx)
    assert False, "Unreachable"
```

jhall1-godaddy: Oh actually, I am able to get around this if I improve the typehint of `util.fail` as follows:

```py
def fail(msg: str, ctx: click.Context, code: int = 1) -> typing.NoReturn:
    ...
```

jhall1-godaddy:
Changed the return type of `util.fail` to `typing.NoReturn` which allowed me to remove the assert.

Added 2 flag options to the command:
`--update-configuration` and `--no-update-configuration` - Default is to update the config.
`--remove-duplicates` and `--no-remove-duplicates` - Default is to remove duplicates.

In the case of removing duplicates, the first instance of the signature is kept and the rest are discarded.
Added a paragraph to `upgrading3.rst` regarding this command and its use case.

rbailey-godaddy (approved):
This is really great work. Many thanks for the contribution, which makes life better for both users and developers!

sushantmimani (approved):
Good stuff!

Issue #375, opened 2022-08-04
"Password in URL" regex ignores matches where username is longer than 20 characters OR password is longer than 20 characters OR url is longer than 100 characters


nkaraliova-godaddy:
## 🐛 Bug Report

  While migrating repositories to GHEC, I had encountered examples where tartufo scan missed credentials that should have been reported as "Password in URL" issue type.
  Upon closer examination, it appears that ["Password in URL" regex ]( https://github.com/godaddy/tartufo/blob/main/tartufo/data/default_regexes.json#L29)is currently failing to report matches  where username string length  exceeds 20 characters OR password string length exceeds 20 characters OR url string length exceeds 100 characters.
  
## To Reproduce

  To test, add one of the following examples to the test repo and run tartufo scan against it:
  Examples:
    https://usernamewithmorethan20chars:somepassword@foo.bar/
    https://user:reallyreallylongpassword123@foo.bar/
    https://user:password@really.long.fqdn.port.and.path/that/exceeds/100/characters/in/length/

## Expected Behavior

  The examples above should be getting reported in Tartufo scan results as the following issue type:
        Reason: Regular Expression Match
        Detail: Password in URL

## Code Example

 The faulty regex is defined here:
 https://github.com/godaddy/tartufo/blob/main/tartufo/data/default_regexes.json#L29
## Environment

Tested with tartufo version 3.1.2

Suggested fix: #376

rbailey-godaddy:
Fixed by #376

PR #376, opened 2022-08-05
Update default_regexes.json to expand Password in URL detection

afammartino-godaddy:

* [ ] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
#375 

## What's new?
Update the "Password in URL" default_regexes.json to allow for usernames/passwords of length 40 and remove the requirement of an fqdn/port/location/query_params of length less than 100.

## Notes
I would like additional help/guidance on testing the performance impact of this change. I did some runtime testing and saw minimal increase in cpu time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+v3.2.2 - 5 August 2022
+----------------------
+
+Bug fixes:
+* [#375](https://github.com/godaddy/tartufo/pull/376) - Update the "Password in URL" default_regexes.json to identify the following:
+  * usernames of lengths between 3-40
+  * passwords of length between 3-40 
+  * URL domain name, port, path, query parameters, and fragments of any length
+
 v3.2.1 - 20 July 2022
 ----------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-v3.2.2 - 5 August 2022
+Unreleased
 ----------------------
 
 Bug fixes:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ maintainers = ["GoDaddy <oss@godaddy.com>"]
 name = "tartufo"
 readme = "README.md"
 repository = "https://github.com/godaddy/tartufo/"
-version = "3.2.1"
+version = "3.2.2"
 
 [tool.poetry.scripts]
 tartufo = "tartufo.cli:main"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ maintainers = ["GoDaddy <oss@godaddy.com>"]
 name = "tartufo"
 readme = "README.md"
 repository = "https://github.com/godaddy/tartufo/"
-version = "3.2.2"
+version = "3.2.1"
 
 [tool.poetry.scripts]
 tartufo = "tartufo.cli:main"
--- a/tartufo/data/default_regexes.json
+++ b/tartufo/data/default_regexes.json
@@ -26,7 +26,7 @@
   "Heroku API Key": "[hH][eE][rR][oO][kK][uU].*[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}",
   "MailChimp API Key": "[0-9a-f]{32}-us[0-9]{1,2}",
   "Mailgun API Key": "key-[0-9a-zA-Z]{32}",
-  "Password in URL": "[a-zA-Z]{3,10}://[^/\\s:@]{3,20}:[^/\\s:@]{3,20}@.{1,100}[\"'\\s]",
+  "Password in URL": "[a-zA-Z]{3,10}://[^/\\s:@]{3,40}:[^/\\s:@]{3,40}@[^\"'\\s\\n]+[\"'\\s]",
   "PayPal Braintree Access Token": "access_token\\$production\\$[0-9a-z]{16}\\$[0-9a-f]{32}",
   "Picatic API Key": "sk_live_[0-9a-z]{32}",
   "Slack Webhook": "https://hooks\\.slack\\.com/services/T[a-zA-Z0-9_]{8}/B[a-zA-Z0-9_]{8}/[a-zA-Z0-9_]{24}",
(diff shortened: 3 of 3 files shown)


tartufo/data/default_regexes.json
```diff
@@ -26,7 +26,7 @@
   "Heroku API Key": "[hH][eE][rR][oO][kK][uU].*[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}",
   "MailChimp API Key": "[0-9a-f]{32}-us[0-9]{1,2}",
   "Mailgun API Key": "key-[0-9a-zA-Z]{32}",
-  "Password in URL": "[a-zA-Z]{3,10}://[^/\\s:@]{3,20}:[^/\\s:@]{3,20}@.{1,100}[\"'\\s]",
+  "Password in URL": "[a-zA-Z]{3,10}://[^/\\s:@]{3,40}:[^/\\s:@]{3,40}@[^\"'\\s\\n]+[\"'\\s]",
```
rbailey-godaddy: I'm staring at the latter part of this expression and trying to work my brain around it, unsuccessfully... ;)

What we used to have was a host+path part of 1-100 characters followed by a quotation mark or whitespace. At the very least I could agree it's not uncommon to see longer URLs (longer than 100 characters)... okay and as I am typing this I see it's now any amount of stuff that's *NOT* the ending delimiter (or a newline), followed by the ending delimiter.

Two comments here. First, you can't see it looking here, but the input blobs get picked apart a line at a time so you'll never find a newline `\n` in the middle of any of the search string. Second, I think you could write this more clearly using a non-greedy operator. I would suggest trying `.+?` instead of `.{1,100}` -- this (with respect to the latter part of the expression) will match any number of characters up until the first quotation mark or whitespace character.

Structurally, now that I'm thinking about it, this still seems problematic:
* If you have quotes, they usually come in pairs -- I guess this logic is implicitly assuming the URL might be quoted (and that we are looking for the close quote that matches the leading quote ahead of the beginning of the URL). If we thought the URL itself might contain quotation marks, this would stop short, but I don't think that will be a problem in practice.
* If your URL is not quoted at all and appears at the end of the line (perhaps `URL=https://user:password@example.com` in a shell script) then neither the original nor the new expression will match.

So what's probably needed is an expression that stops at a quotation mark, whitespace, or the end of the line. For this purpose, I think your expression is better than a non-greedy expression, because it allows us to simply drop the closing delimiter:
```suggestion
  "Password in URL": "[a-zA-Z]{3,10}://[^/\\s:@]{3,40}:[^/\\s:@]{3,40}@[^\"'\\s]+",
```

rbailey-godaddy (changes requested):
I like what you did, but I believe I spotted another problem while I was trying to understand it. Does the comment below make sense?

tartufo/data/default_regexes.json
rbailey-godaddy: I forgot to remove the unnecessary `\n` when I first posted this, even though I commented on it earlier. I updated the suggestion to remove it.

afammartino-godaddy: Hey Scott, I see what you mean. In the case that a URL does not currently end in a quote or whitespace we'll miss. I didn't catch that my first pass on this.
I like your suggested change, but before making it I wanted to bring up something. As we search through a line with this pattern, if we get to the @ sign portion of the regex then we have found a URL that matches the important part of the pattern. With that said, what are your thoughts on this?
`[a-zA-Z]{3,10}://[^/\s:@]{3,40}:[^/\s:@]{3,40}@.+`
Does it matter if we report anything past the URL? I still think your suggestion is the best approach, just asking what is the methodology behind these patterns? 

One additional thing to consider is that we are matching templatized strings that don't contain spaces, or strings that contain ENV variables. I think this could be reported separately from bug #375 

Lastly, I'm not sure if writing tests for this data is appropriate for the project as the default_regexes can be changed by the user, but do you think a smoke-test suite should be put together for the data? One that is run on change of data or publishing a release? Here is some example test data.
```
# Patterns that should match:
https://user:password@example.com/foo/bar/gdg
"https://"USER":"password"@example.com/foo/bar/gdg'
URL=https://user:password@example.com
URL = https://user:password@example.com
url: https://user:password@example.com/foo/bar/gdg
url: "https://user:password@example.com/foo/bar/gdg"
url: 'https://user:password@example.com/foo/bar/gdg'

# ENV file matches along with Jinja2/ERB style templating:
url: 'https://${USERNAME}:${PASSWORD}@example.com/foo/bar/gdg'
url: 'https://<%=USERNAME%>:<%=PASSWORD%->@example.com/foo/bar/gdg'
url: 'https://{%USERNAME%}:<%=PASSWORD%->@example.com/foo/bar/gdg'
```

Thanks for the feedback!

afammartino-godaddy:
Hey just wondering if anyone else had some thoughts on how to handle tests for these regex changes?

rbailey-godaddy (approved):

sushantmimani (approved):

pyproject.toml
```diff
@@ -32,7 +32,7 @@ maintainers = ["GoDaddy <oss@godaddy.com>"]
 name = "tartufo"
 readme = "README.md"
 repository = "https://github.com/godaddy/tartufo/"
-version = "3.2.1"
+version = "3.2.2"
```
emayuri-godaddy: I believe you need not update the version of tartufo as part of bug fix or implementation PR. This will be updated when a release is created

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+v3.2.2 - 5 August 2022
```
emayuri-godaddy: Also this should be 
```suggestion
Unreleased
```

emayuri-godaddy (changes requested):

pyproject.toml
rbailey-godaddy: Good catch, Esha! Anthony, can you revert the two version changes (here and below) she called out?

afammartino-godaddy: Thank you for the review. I've reverted the change.

rbailey-godaddy (approved):

emayuri-godaddy (approved):

PR #401, opened 2022-11-18
feat(output): add report output format

dclayton-godaddy:

* [x] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [x] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?

- added report format to output a reviewer-friendly report
- fix Docker build

```text
$ tartufo --output-format report scan-local-repo .
Tartufo Scan Results (Time: 2022-11-09T16:24:34.012693)

Configuration:
  version:             3.2.1
  entropy:             Enabled
    sensitivity: 75
  regex:               Enabled

Excluded paths:
  package-lock.json: Autogenerated package lock file
  cicd.yaml: cert thumbprint
  README.md: readme example

Excluded signatures:
  a32cedf592e11c8d2efd06e4bd7906b340b64cb2f4b1cbb925fe908042d5d7bb: simple test file

Excluded entropy patterns:
  thumbprint:\s{1}[a-zA-Z0-9]{40} (path=cicd.yaml, scope=line, type=search): cert thumbprint
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,14 @@ ENV PIP_DEFAULT_TIMEOUT=100 \
     PIP_NO_CACHE_DIR=1 \
     POETRY_VERSION=1.1.12
 
+RUN apt-get update && apt-get upgrade -y && apt-get install make cmake gcc libssl-dev wget -y
+RUN wget https://github.com/libgit2/libgit2/archive/refs/tags/v1.4.0.tar.gz -O libgit2-1.4.0.tar.gz \
+    && tar xzf libgit2-1.4.0.tar.gz \
+    && cd libgit2-1.4.0/ \
+    && cmake . \
+    && make \
+    && make install
+RUN apt-get install libffi-dev -y
 RUN pip --no-cache-dir install "poetry==$POETRY_VERSION"
 RUN python -m venv /venv
 
--- a/tartufo/cli.py
+++ b/tartufo/cli.py
@@ -122,6 +122,7 @@ class TartufoCLI(click.MultiCommand):
             types.OutputFormat.Json.value,
             types.OutputFormat.Compact.value,
             types.OutputFormat.Text.value,
+            types.OutputFormat.Report.value,
         ]
     ),
     default="text",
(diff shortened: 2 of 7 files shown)


sushantmimani (approved):

Dockerfile
```diff
@@ -9,6 +9,14 @@ ENV PIP_DEFAULT_TIMEOUT=100 \
...
     POETRY_VERSION=1.1.12
 
+RUN apt-get update && apt-get upgrade -y && apt-get install make cmake gcc libssl-dev wget -y
+RUN wget https://github.com/libgit2/libgit2/archive/refs/tags/v1.4.0.tar.gz -O libgit2-1.4.0.tar.gz \
+    && tar xzf libgit2-1.4.0.tar.gz \
+    && cd libgit2-1.4.0/ \
+    && cmake . \
+    && make \
+    && make install
+RUN apt-get install libffi-dev -y
```
rbailey-godaddy, line 19: I wouldn't spend more than 10 minutes on this, but I was grappling with the problem independently this morning and have two observations:

1. The error is due to being unable to install binary wheels into the container. This approach works but isn't strictly required.
2. My alternate solution was to bump pygit2 from ^1.9.0 to ^1.10.0 for python versions 3.7 and later. However, be careful because I found that when I rebuilt poetry.lock with a new poetry (1.2.2) all of the unit testing for python 3.6 (which is forced to use the old poetry 1.1.x) broke because of a change in behavior surrounding setuptools support between poetry 1.1 and poetry 1.2.

I cannot explain why 1.9.x no longer works, but perhaps the base container python version bumped and 1.9.x doesn't have the needed content any longer? What you have is okay and has the virtue of working, but it's faintly annoying that we have to do it in the first place. See #403 for more ranting on this subject.

rbailey-godaddy (approved):
One comment on the docker build, but maybe it's better to leave this as you propose now and circle back with #403 after we get #400 merged.

Dockerfile
dclayton-godaddy: Ya, the build was failing for quite some time. This at least gets it to work.

PR #421, opened 2023-01-04
feat: add progress flag to git scans

dclayton-godaddy:

* [x] Tests (if applicable)
* [ ] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [x] Refactoring (no functional changes, no api changes)
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
- add `--progress` to git scans to display overall and branch progress
- only call is_binary if chunk is not delete


sjacoby-godaddy (approved):

tartufo/scanner.py
```diff
@@ -871,64 +919,38 @@ def chunks(self) -> Generator[types.Chunk, None, None]:
...
                 commits = [self._repo.get(self._repo.head.target)]
             else:
                 branch = self._repo.branches.get(branch_name)
                 try:
-                    commits = self._repo.walk(
-                        branch.resolve().target, pygit2.GIT_SORT_TOPOLOGICAL
+                    commits = list(
+                        self._repo.walk(
+                            branch.resolve().target, pygit2.GIT_SORT_TOPOLOGICAL
+                        )
```
rbailey-godaddy: I'm guessing `self._repo.walk()` is returning some sort of iterator here, and you're converting it to a list so that you can take the length of it below for use in constructing the progress information. I worry that this may impose significant overhead (both in memory consumption and latency) on scans of large repositories.

Is it possible to defer the list conversion and then perform it only in the case `self.git_options.progress` is true?

rbailey-godaddy (changes requested):
I like this, but I'd like it better if we tweaked it to eliminate the overhead of converting `commits` to a list when a progress bar is not requested. That seems like it should be possible...

tartufo/scanner.py
dclayton-godaddy: I agree. Updated. Unfortunately I need to convert to list for the progressbar, otherwise it wont show progress.

dclayton-godaddy: Also rebased and addressed merge conflicts.

rbailey-godaddy (approved):
Thanks for the fix - I like it.

sushantmimani (approved):

PR #458, opened 2023-03-08
add exclude-regex-patterns

jpaskhay:
This mimics the functionality of the `exclude-entropy-patterns` option, but for regex scans. It is heavily based off #192 and other subsequent work around that feature. This will help reduce false positives such as env variables for user info auth in URLs.

Note that the `scope` is forced to `line` and not configurable due to the way the regex scan is done compared to the entropy scan.

A few unrelated formatting changes were necessary for `pre-commit` linters to pass.

I ran into issues trying to get the development environment setup properly (mainly poetry version differences). The new tests that were added passed, but I was running into 2 unrelated test failures -- not sure if related to the dev setup issues.

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [x] Code style update (formatting, local variables)
* [ ] Build related changes
* [ ] CI related changes
* [x] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## What's new?
- add `exclude-regex-patterns` option

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -203,6 +203,53 @@ match-type   No       String ("search" or "match")  Whether to perform a `search
 scope        No       String ("word" or "line")    Whether to match against the current word or full line of text
 ============ ======== ============================ ==============================================================
 
+.. regex-exclusion-patterns:
+
+Regex Exclusion Patterns
+++++++++++++++++++++++++++
+
+Regex scans can produce false positive matches such as environment variables in
+URLs. To avoid these false positives, you can use the
+``exclude-regex-patterns`` configuration option. These patterns will be
+applied to and matched against any strings flagged by regex pattern checks. As
+above, this directive utilizes an `array of tables`_, enabling two forms:
+
+Option 1:
+
+.. code-block:: toml
+
+    [tool.tartufo]
+    exclude-regex-patterns = [
+        {path-pattern = 'products_.*\.txt', pattern = '^SK[\d]{16,32}$', reason = 'SKU pattern that resembles Twilio API Key'},
+        {path-pattern = '\.github/workflows/.*\.yaml', pattern = 'https://\${\S+}:\${\S+}@\S+', reason = 'URL with env variables for auth'},
+    ]
+
+Option 2:
+
+.. code-block:: toml
+
+    [[tool.tartufo.exclude-regex-patterns]]
+    path-pattern = 'products_.*\.txt'
+    pattern = '^SK[\d]{16,32}$'
+    reason = 'SKU pattern that resembles Twilio API Key'
+
+    [[tool.tartufo.exclude-regex-patterns]]
+    path-pattern = '\.github/workflows/.*\.yaml'
+    pattern = 'https://\${\S+}:\${\S+}@\S+'
+    reason = 'URL with env variables for auth'
+
+
+There are 4 relevant keys for this directive, as described below.
+
+============ ======== ============================ ==============================================================
+Key          Required Value                        Description
+============ ======== ============================ ==============================================================
+pattern      Yes      Regular expression           The pattern used to check against the match
+path-pattern No       Regular expression           A pattern to specify to what files the exclusion will apply
+reason       No       String                       A plaintext reason the exclusion has been added
+match-type   No       String ("search" or "match")  Whether to perform a `search or match`_ regex operation
+============ ======== ============================ ==============================================================
+
 .. _TOML: https://toml.io/
 .. _array of tables: https://toml.io/en/v1.0.0#array-of-tables
 .. _search or match: https://docs.python.org/3/library/re.html#search-vs-match
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -206,7 +206,7 @@ scope        No       String ("word" or "line")    Whether to match against the 
 .. regex-exclusion-patterns:
 
 Regex Exclusion Patterns
-++++++++++++++++++++++++++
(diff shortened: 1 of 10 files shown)


jpaskhay:
welp, guess that unit test failure wasn't related to my dev environment 😄 . will troubleshoot a bit more and hopefully have a fix soon

docs/source/configuration.rst
```diff
@@ -203,6 +203,53 @@ match-type   No       String ("search" or "match")  Whether to perform a `search
 scope        No       String ("word" or "line")    Whether to match against the current word or full line of text
 ============ ======== ============================ ==============================================================
 
+.. regex-exclusion-patterns:
+
+Regex Exclusion Patterns
+++++++++++++++++++++++++++
```
rbailey-godaddy:
```suggestion
Regex Exclusion Patterns
++++++++++++++++++++++++
```
I think the underline length needs to match the header

docs/source/configuration.rst
```diff
@@ -203,6 +203,53 @@ match-type   No       String ("search" or "match")  Whether to perform a `search
...
+
+============ ======== ============================ ==============================================================
+Key          Required Value                        Description
+============ ======== ============================ ==============================================================
+pattern      Yes      Regular expression           The pattern used to check against the match
+path-pattern No       Regular expression           A pattern to specify to what files the exclusion will apply
+reason       No       String                       A plaintext reason the exclusion has been added
+match-type   No       String ("search" or "match")  Whether to perform a `search or match`_ regex operation
+============ ======== ============================ ==============================================================
+
```
rbailey-godaddy, line 317: You addressed this in the PR description, but it would be worth throwing in a line something like "The pattern is always tested against the entire line." (Or any other words that would explain why `scope` is a thing just above for entropy exclusions, but isn't here.)

tartufo/config.py
```diff
@@ -278,12 +278,15 @@ def compile_rules(patterns: Iterable[Dict[str, str]]) -> List[Rule]:
...
-            ) from exc
+        if exclude_type != "regex":
+            try:
+                scope = Scope(pattern.get("scope", Scope.Line.value))
+            except ValueError as exc:
+                raise ConfigException(
+                    f"Invalid value for scope: {pattern.get('scope')}"
+                ) from exc
+        else:
+            scope = Scope.Line
```
rbailey-godaddy: This would be more direct this way:
```suggestion
        if exclude_type == "regex":
            # regex exclusions always have line scope
            scope = Scope.Line
        else:
            # entropy exclusions can specify scope
            try:
                scope = Scope(pattern.get("scope", Scope.Line.value))
            except ValueError as exc:
                raise ConfigException(
                    f"Invalid value for scope: {pattern.get('scope')}"
                ) from exc
```

tartufo/util.py
```diff
@@ -242,7 +257,7 @@ def _style_func(msg: str, *_: Any, **__: Any) -> str:
     style_ok = style_error = style_warning = partial(_style_func)
 
 
-def fail(msg: str, ctx: click.Context, code: int = 1) -> NoReturn:
+def fail(msg: str, ctx: click.Context, code: int = 1) -> None:
```
rbailey-godaddy: I haven't gotten to looking at the unit test failures yet, but I was surprised to see this return type change. What is the motivation for this? Isn't it still true that `fail()` never returns control to the caller?

rbailey-godaddy (changes requested):
This generally looks pretty good - I have some nitpicking below and then there's the question of the failing CI tests, which we'll look at next...

docs/source/configuration.rst
jpaskhay: sure. will do

tartufo/config.py
jpaskhay: good catch, agree here. ended up focusing on the unit tests that i didn't self-scrutinize the code changes as much as i should have

tartufo/util.py
jpaskhay: i was getting a `mypy` failure on the pre-commit hook with:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1
tartufo/util.py:261: error: Implicit return in function which does not return
```
i can try reverting this to see if it complains in CICD. this change as is has caused another side-effect linting issue, which i was surprised wasn't flagged locally in pre-commit

rbailey-godaddy: I think the mypy failure you're getting now is due to the return-value change:

```
tartufo/commands/update_signatures.py:273: error: Incompatible return value type (got "Optional[GitRepoScanner]", expected "GitRepoScanner")  [return-value]
```
It used to be that `scanner` was known not to be `None` by L273 because of the test at L250 combined with the knowledge that control would not return from the call to `util.fail()` on L252. Now mypy assumes control will return and that consequently `scanner` could still be `None` -- thus the error.

jpaskhay: yup, exactly. was just posting the previous mypy failure which led me to the `NoReturn` -> `None` change. wrapping up wordsmithing the documentation change and i'll push up a revert for this line along with it.

jpaskhay:
i'm at a loss on the unit test failure. i did some troubleshooting locally, and the `blake2s` isn't getting mocked, so the real thing is getting called

the way it is getting patched seems proper (same w/ the various other patches in `test_util.py`). not seeing anything obvious that would've changed that in the change set 🤔 . even tried adding a `importlib.reload(util)`

rbailey-godaddy (approved):

emayuri-godaddy (approved):

PR #468, opened 2023-03-21
Bugfix: multiple configuration files not processed

### LINKED ISSUE #467: Only the last of multiple configuration files is processed
## 🐛 Bug Report

If tartufo is executed using multiple `--config` options, only the last-specified configuration file is processed.

## To Reproduce

Using the tartufo repository itself as an example:

1. Copy the tartufo stuff out of `pyproject.toml` into a new `foo.toml` (leave all of the existing stuff alone so that what you are trying to look at doesn't get washed out with findings)
2. Add a new `exclude-path-patterns` entry such as `{path-pattern = 'foo', reason = 'Testing'},` to `foo.toml`
3. Duplicate `foo.toml` to `bar.toml`, and change that new entry from `foo` to `bar`
4. Run `tartufo -v --config foo.toml --config bar.toml scan-local-repo .`

You can see in the output that the excluded paths list includes bar but not foo. Switch the order of the files and you get opposite results.

Interestingly, if you add a short `exclude-entropy-patterns` in `foo.toml`, for example:

```toml
exclude-entropy-patterns = [
    {path-pattern = 'foo', pattern = 'whatever', reason = 'Testing'},
]
```

this will get erased by `bar.toml` (even though `bar.toml` doesn't have an `exclude-entropy-patterns` in it), so it really is a case of "all but the last --config have no effect"

## Expected Behavior

The multiple humans I have discussed this issue with expect that configuration files should be cumulative.

* List-valued parameters, such as `exclude-path-patterns` and `exclude-entropy-patterns`, should be merged
* Single-valued parameters, such as `cleanup` and `entropy`, should be determined by the last-specified configuration

## Environment

I am reporting this problem against tartufo 4.0.1, but I believe this behavior exists in all previous versions of tartufo.

rbailey-godaddy:

* [X] Tests (if applicable)
* [X] Documentation (if applicable)
* [X] Changelog entry

## PR Type

* [X] Bugfix
* [ ] Feature
* [ ] Build related changes
* [ ] CI related changes
* [X] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

This is arguable but I believe "backward compatible" best describes the spirit of the change.

Behavior has changed, but the new behavior matches what users thought was the behavior all along (i.e. if you specify multiple configuration files, all specified configuration files will be used). The `read_project_toml()` function still returns a string, but in the event multiple files were specified, that string is a comma-delimited list of filenames instead of a single filename. However, no in-tree code ever looks at this return value (except for some unit tests).

## Issue Linking

fixes #467 

## What's new?
Several configuration file processing bugs have been fixed.

* When the `--config` parameter appears multiple times, tartufo now processes each specified file in order; previously it ignored all but the last file. The documentation was silent on whether this was a valid thing or not, but no warning or error was generated and users were doing this and believing it would do something useful.
* When running `scan-folder`, an oversight caused the target folder not to be scanned for configuration files. Now tartufo will scan for configuration files, the same way it does for `scan-local-repo` and `scan-remote-repo`
* The same problem occurred for `pre-commit`.
* The logic that was supposed to avoid reloading a configuration file was horribly broken before, and it was even more broken by added support for multiple `--config` options. This has been fixed.
* To reduce the chances of screwing this up in the future, Configuration processing has been pulled into a `load_config()` method and all of the open-coded stuff -- typically in `load_repo()` although it had nothing to do with repository loading -- has been replaced with calls to the new method

Additionally, merging of data from multiple files is improved. (This wasn't really a factor before, because tartufo never looked at more than one user-specified configuration file.) Single-valued attributes are overwritten (so the last configuration file that specifies a setting wins) and list-valued attributes are concatenated (generally aligning to the expected behavior). The documentation has been updated to describe this, and unit tests have been added to verify that's really what the code does.

The documentation updates also include greatly expanded material on how configuration files are detected, since that appears to be a point of enduring confusion among both users and maintainers. ;-)

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -8,28 +8,93 @@ unwieldy and lead to an overly cumbersome command. It also becomes difficult to
 reliably reproduce the same command in all environments when done this way.
 
 To help with these problems, ``tartufo`` can also be configured by way of a
-configuration file! You can `tell tartufo what config file to use
-<usage.html#cmdoption-tartufo-config>`__, or, it will automatically discover one
-for you. Starting in the current working directory, and traversing backward up
-the directory tree, it will search for both a ``tartufo.toml`` and a
-``pyproject.toml``. The latter is searched for as a matter of convenience for
-Python projects, such as ``tartufo`` itself. For an example of the tree
-traversal, let's say you running ``tartufo`` from the directory
-``/home/my_user/projects/my_project``. ``tartufo`` will look for the
-configuration files first in this directory, then in ``/home/my_user/projects/``,
-then in ``/home/my_user``, etc. The search stops when the first non-empty file
-is encountered.
+configuration file!
 
-If multiple configuration files are specified, they will be located and processed
-in order. When conflicting directives are provided in different files, the value
-in the last file processed takes precedence. List-valued directives (such as
+You can `tell tartufo what config file to use
+<usage.html#cmdoption-tartufo-config>`__ by specifying one or more configuration
+files on the command line. Each file is located and processed in order. When
+conflicting directives are provided in different files, the value in the last
+file processed takes precedence. List-valued directives (such as
 ``exclude-path-patterns``) that are present in multiple files are concatenated.
 
-Within these files, ``tartufo`` will look for a section labeled
-``[tool.tartufo]`` to find its configuration, and will load all items from there
-just as though they had been specified on the command line. This file must be
+.. _configuration-discovery:
+
+Configuration Discovery
+-----------------------
+
+``tartufo`` uses two additional heuristics to locate specified configuration files.
+First, the current working directory is used as a base for relative filenames.
+Second, if the specified file does not exist in the specified directory, ``tartufo``
+will search upwards in the directory hierarchy looking for a file with the same
+name.
+
+Additionally, ``tartufo`` will look for a configuration file in the scan target
+repository or folder. It looks first for ``tartufo.toml``, and if that does not
+exist, then ``pyproject.toml``. The latter is searched for as a matter of
+convenience for Python projects, such as ``tartufo`` itself. This file must
+exist in the repository root directory (or target folder). The discovered
+configuration file will be processed after configuration files specified on the
+command line, unless it was also specified on the command line -- in which case,
+it will not be read a second time.
+
+Therefore, while it normally a target's configuration file will override settings
+specified using ``--config`` on the command line, this behavior can be overridden
+by specifying the target's configuration explicitly first, and then the desired
+"master" configuration second.
+
+Consider:
+
+.. code-block:: shell
+
+   tartufo --config myconfig.toml scan-local-repo .
+
+``tartufo`` will look for ``myconfig.toml`` in the current directory, and then
+in the parent directory if it does not exist in ``.``, etc., and then look for
+either ``tartufo.toml`` or (if not found) ``pyproject.toml`` in the current
+directory (only) because that is the target of the scan. Directives in, say,
+``tartufo.toml`` would supersede settings in ``myconfig.toml``.
+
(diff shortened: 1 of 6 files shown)


docs/source/configuration.rst
```diff
@@ -8,27 +8,98 @@ unwieldy and lead to an overly cumbersome command. It also becomes difficult to
...
+convenience for Python projects, such as ``tartufo`` itself. This file must
+exist in the repository root directory (or target folder). The discovered
+configuration file will be processed after configuration files specified on the
+command line, unless it was also specified on the command line -- in which case,
+it will not be read a second time.
+
+Therefore, while it normally a target's configuration file will override settings
+specified using ``--config`` on the command line, this behavior can be overridden
+by specifying the target's configuration explicitly first, and then the desired
+"master" configuration second.
```
sushantmimani: I think this paragraph might be missing a few words.

docs/source/configuration.rst
```diff
@@ -8,27 +8,98 @@ unwieldy and lead to an overly cumbersome command. It also becomes difficult to
...
+was found).
+
+If ``tartufo.toml`` is not present but exists in the parent directory, then the
+effect is that ``../tartufo.toml`` is read first, ``myconfig.toml`` is read next,
+and ``pyproject.toml`` (if it exists) would be processed last because it is in
+the scan target and ``tartufo.toml`` does not exist in the scan root.
+
+Note this behavior is not applicable to remote repository scans, because the
+remote repository will be cloned to a scratch directory and neither that directory
+nor the configuration files within it are available to specify with ``-config``.
```
sushantmimani: We should probably make the `note` bold to make it stand out?

emayuri-godaddy (approved):
LGTM

docs/source/configuration.rst
rbailey-godaddy: Actually, on re-read, I think it has a word or two too many :) -- let me try to clean up that and the one below

sushantmimani (approved):
🚀

docs/source/configuration.rst
```diff
@@ -8,27 +8,98 @@ unwieldy and lead to an overly cumbersome command. It also becomes difficult to
...
+For purposes of this scan, empty files are considered not to exist.
+
+However:
+
+.. code-block:: shell
+
+   tartufo --config tartufo.toml --config myconfig.toml scan-local-repo .
+
+will cause ``tartufo`` to read ``tartufo.toml`` in the current directory
+(assuming it exists, and in parent directories otherwise), and then ``myconfig.toml``
```
jsimmons1-godaddy, line 66: `(assuming it exists, and in parent directories otherwise)` is this accurate? Unless the behavior changed in this PR (which I don't see, but is quite possible), it will not traverse upwards in this case, because `tartufo.toml` gets turned into an absolute path to the file.

rbailey-godaddy: I believe this is still correct (although we plan a PR before the next release that will unilaterally remove the traversal logic, which we believe is effectively unused and won't be missed).

If you pass a relative path via `--config`, that is sent to `config.read_pyproject_toml()`, which sends it (unmodified) to `config.load_config_from_path()` (with `config_path` set to `.` and  `traverse` implicitly `True`).

If it can't find a file, `load_config_from_path()` recursively calls itself, except this time `config_path` is set to the parent of the original path (i.e. `../`) and repeats this until you reach the root directory, at which point `config_path.parent` matches `config_path` and the loop terminates.

jsimmons1-godaddy: Okay, thanks. Yeah it sound like it will become immaterial soon, but just for completion: my understanding of the current/old behavior is: If you pass a relative path via `--config` , the click library transforms it immediately into an absolute path. The rest appears correct, except it is effectively unable to traverse due to that transformation that has already been done.

PR #473, opened 2023-03-23
Setting up new flag --no-target-config to skip use of local config files if required

### LINKED ISSUE #469: Tartufo looks for config in current directory even when using --config
## 🐛 Bug Report

If tartufo is executed `tartufo --config /path/to/config scan-local-repo .` It still imports any tartufo config present in the current directory in addition to the config specified in `/path/to/config`

## To Reproduce

Using the tartufo repository itself as an example:

1. Copy some of the tartufo stuff out of `pyproject.toml` into a new `foo.toml` (leave all of the existing stuff alone so that what you are trying to look at doesn't get washed out with findings)
2. Run `tartufo -v --config foo.toml scan-local-repo .`

## Expected Behavior

It should only use the config in `foo.toml`

## Environment

Tartufo v4.0.1

emayuri-godaddy:

* [x] Tests (if applicable)
* [x] Changelog entry

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [ ] No (breaking changes)

## Issue Linking
fixes #469

## What's new?
- Adding a new flag `--no-target-config` to skip processing of repository's config file if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ vX.X.X - Mar 3 2023
 --------------------
 
 Features:
+* [#473](https://github.com/godaddy/tartufo/pull/473) - Introduces new flag `--no-local-config` to skip processing of local config files if needed.
 * [#455](https://github.com/godaddy/tartufo/pull/455) - Update documentation to fix incorrect wording
 
 Bug fixes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ vX.X.X - Mar 3 2023
 --------------------
 
 Features:
-* [#473](https://github.com/godaddy/tartufo/pull/473) - Introduces new flag `--local-config/--no-local-config` to process or skip using local config files.
+* [#473](https://github.com/godaddy/tartufo/pull/473) - Introduces new flag `-target-config/--no-target-config` to process or skip using config file in the repository being scanned
 * [#455](https://github.com/godaddy/tartufo/pull/455) - Update documentation to fix incorrect wording
 
 Bug fixes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ vX.X.X - Mar 3 2023
 --------------------
 
 Features:
-* [#473](https://github.com/godaddy/tartufo/pull/473) - Introduces new flag `--no-local-config` to skip processing of local config files if needed.
+* [#473](https://github.com/godaddy/tartufo/pull/473) - Introduces new flag `--local-config/--no-local-config` to process or skip using local config files.
 * [#455](https://github.com/godaddy/tartufo/pull/455) - Update documentation to fix incorrect wording
 
 Bug fixes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ vX.X.X - Mar 3 2023
 Features:
 * [#455](https://github.com/godaddy/tartufo/pull/455) - Update documentation to fix incorrect wording
 
+Bug fixes:
+* [#467](https://github.com/godaddy/tartufo/issues/467) - Multiple fixes to configuration
+  file processing:
+  - If multiple configuration files were specified, only the last was processed
+    and no error or warning was generated. Now files are processed in order.
+  - When multiple configuration files are specified, list-valued parameters are
+    concatenated and single-valued parameters are overwritten by the last file
+    that defines them.
+  - Configuration files located in the target of a `scan-folder` operation were
+    ignored; now they are located and processed in the same manner as for a
+    `scan-local-repo` or `scan-remote-repo` operation.
+
 v4.0.1 - Mar 1 2023
 --------------------
 
--- a/README.md
+++ b/README.md
@@ -87,10 +87,11 @@ Options:
                                   specified multiple times.
   --config FILE                   Read configuration from specified file.
                                   [default: tartufo.toml]
-  --local-config / --no-local-config BOOL
-                                  Enable or Disable processing of local config file
+  --target-config/--no-target-config BOOL
+                                  Enable or Disable processing of config file in the
+                                  repository being scanned
                                   i.e. config files like tartufo.toml or pyproject.toml
-                                  setup in local working directory
+                                  setup in the repository being scanned
                                   [default: local-config]
   -q, --quiet / --no-quiet        Quiet mode. No outputs are reported if the
                                   scan is successful and doesn't find any
--- a/README.md
+++ b/README.md
@@ -87,6 +87,10 @@ Options:
                                   specified multiple times.
   --config FILE                   Read configuration from specified file.
                                   [default: tartufo.toml]
+  --no-local-config BOOL          Skip processing of local config file
+                                  i.e. config files like tartufo.toml or pyproject.toml
+                                  setup in local working directory
+                                  [default: False]
   -q, --quiet / --no-quiet        Quiet mode. No outputs are reported if the
                                   scan is successful and doesn't find any
                                   issues
--- a/README.md
+++ b/README.md
@@ -87,10 +87,11 @@ Options:
                                   specified multiple times.
   --config FILE                   Read configuration from specified file.
                                   [default: tartufo.toml]
-  --no-local-config BOOL          Skip processing of local config file
+  --local-config / --no-local-config BOOL
+                                  Enable or Disable processing of local config file
                                   i.e. config files like tartufo.toml or pyproject.toml
                                   setup in local working directory
-                                  [default: False]
+                                  [default: local-config]
   -q, --quiet / --no-quiet        Quiet mode. No outputs are reported if the
                                   scan is successful and doesn't find any
                                   issues
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -8,28 +8,93 @@ unwieldy and lead to an overly cumbersome command. It also becomes difficult to
 reliably reproduce the same command in all environments when done this way.
 
 To help with these problems, ``tartufo`` can also be configured by way of a
-configuration file! You can `tell tartufo what config file to use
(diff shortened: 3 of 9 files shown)


tartufo/cli.py
```diff
@@ -252,6 +252,13 @@ def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Comma
     help="""Enable or disable terminal color. If not provided (default), enabled if
     output is a terminal (TTY).""",
 )
+@click.option(
+    "--no-local-config",
+    is_flag=True,
+    default=False,
+    show_default=True,
+    help="Skipping auto discovery and use of local config file",
+)
```
rbailey-godaddy: This would look more like the other options if you structured this as `--local-config/--no-local-config` with the default being `True` (i.e. `--local-config`) -- look at the immediately following quiet definition for an example.

rbailey-godaddy (changes requested):
I'd like to see this tweaked so the flag is "local_config" with default True; more notes below where the definition is made. Obviously other references to it would change with respect to the name and the sense.

jsimmons1-godaddy:
Which config file exactly does this prevent using?

For example in this case: `tartufo --no-local-config scan-local-repo foo/bar/baz/` , assuming that there is a `tartufo.toml` both in my current directory, and in the `foo/bar/baz` directory, which config is now being skipped that would have been used before?

I ask because looking at #469 (that I believe this is meant to address) , the command is `tartufo --config /path/to/config scan-local-repo .` , it's only using the config file in the current directory because `.` , the current directory, is the repo that scan-local-repo was given. So if this is meant to prevent using the `tartufo.toml` from the target, I think the wording of the flag and documentation should be different.

emayuri-godaddy:
> Which config file exactly does this prevent using?
> 
> For example in this case: `tartufo --no-local-config scan-local-repo foo/bar/baz/` , assuming that there is a `tartufo.toml` both in my current directory, and in the `foo/bar/baz` directory, which config is now being skipped that would have been used before?
> 
> I ask because looking at #469 (that I believe this is meant to address) , the command is `tartufo --config /path/to/config scan-local-repo .` , it's only using the config file in the current directory because `.` , the current directory, is the repo that scan-local-repo was given. So if this is meant to prevent using the `tartufo.toml` from the target, I think the wording of the flag and documentation should be different.

It prevents the config in the working directory that is being scanned.

emayuri-godaddy:
> Which config file exactly does this prevent using?
> 
> For example in this case: `tartufo --no-local-config scan-local-repo foo/bar/baz/` , assuming that there is a `tartufo.toml` both in my current directory, and in the `foo/bar/baz` directory, which config is now being skipped that would have been used before?
> 
> I ask because looking at #469 (that I believe this is meant to address), the command is `tartufo --config /path/to/config scan-local-repo .` , it's only using the config file in the current directory because `.` , the current directory, is the repo that scan-local-repo was given. So if this is meant to prevent using the `tartufo.toml` from the target, I think the wording of the flag and documentation should be different.

It prevents processing of config file in the directory that is being scanned whether it be local or remote
In your example `tartufo --no-local-config scan-local-repo foo/bar/baz/` it should skip processing of toml file in foo/bar/baz. 
I agree the --local/--no-local-config might be a little miss leading. We are planning on re-naming it to --target-config/--no-target-config and updating documentation to be more clear.

README.md
```diff
@@ -87,10 +87,11 @@ Options:
...
   --config FILE                   Read configuration from specified file.
                                   [default: tartufo.toml]
-  --local-config / --no-local-config BOOL
-                                  Enable or Disable processing of local config file
+  --target-config/--no-target-config BOOL
+                                  Enable or Disable processing of config file in the
+                                  repository being scanned
                                   i.e. config files like tartufo.toml or pyproject.toml
-                                  setup in local working directory
+                                  setup in the repository being scanned
```
jsimmons1-godaddy, line 94: Does this flag apply to the `scan-folder` subcommand as well, or only `scan-local-repo` and `scan-remote-repo` ? What about `pre-commit` ? Depending on which it applies to, the wording could potentially be improved here.

jsimmons1-godaddy:
> It prevents processing of config file in the directory that is being scanned whether it be local or remote
> In your example `tartufo --no-local-config scan-local-repo foo/bar/baz/` it should skip processing of toml file in foo/bar/baz.
> I agree the --local/--no-local-config might be a little miss leading. We are planning on re-naming it to --target-config/--no-target-config and updating documentation to be more clear.

Thanks, yeah I find that wording much more clear.

I'm curious if you have thought about there being another (separate) flag that prevents processing the current-directory config as well? For `tartufo --no-target-config scan-local-repo foo/bar/baz/` , one that would skip processing the `tartufo.toml` in the current directory in which this is run (or in parent directories of the current directories if it hasn't found one yet).

emayuri-godaddy:
> 

As far as I understand there are only two ways of processing config files one in the target directory that is being scanned which is enabled by default and now can be disabled by using `--no-target-config` and the other option is to provide config file using `--config` which traverses the directory to find required file. The traversal of the directory when `--config` is provided will be disabled as part of another issue #475

README.md
emayuri-godaddy: It should apply to both `scan-folder` and `pre-commit` along with `scan-local-repo` and `scan-remote-repo`

jsimmons1-godaddy:
> > 
> 
> As far as I understand there are only two ways of processing config files one in the target directory that is being scanned which is enabled by default and now can be disabled by using `--no-target-config` and the other option is to provide config file using `--config` which traverses the directory to find required file. The traversal of the directory when `--config` is provided will be disabled as part of another issue #475

If you don't specify `--config` at all, it will assume you mean `tartufo.toml` or `pyproject.toml` in the current directory, where you're running this command (or parent directories, if it doesn't find either there). Which I don't believe there is currently any way to turn that off, even after this PR.

I have no objections at all with that being addressed in a separate PR though, if it even is something that should have a flag (I'd say yes). Just thought it seemed related enough to mention. (To be clear, I'm just an outside observer anyway, I'm not a maintainer, so even if I did have objections they wouldn't necessarily mean much).

README.md
jsimmons1-godaddy: Okay, I would argue that this should not say `the repository being scanned` then, since it applies to `scan-folder` too. Maybe just `the repository/folder` being scanned?

README.md
```diff
@@ -87,6 +87,12 @@ Options:
                                   specified multiple times.
   --config FILE                   Read configuration from specified file.
                                   [default: tartufo.toml]
+  --target-config/--no-target-config BOOL
```
jsimmons1-godaddy, line 90:
```suggestion
  --target-config/--no-target-config
```

I think this better aligns with the usage documentation for eg `--scan-filenames / --no-scan-filenames` (in that it doesn't imply that a boolean value is to be specified after the argument name)

README.md
```diff
@@ -87,6 +87,12 @@ Options:
...
+                                  Enable or Disable processing of config file in the
+                                  repository being scanned
+                                  i.e. config files like tartufo.toml or pyproject.toml
+                                  setup in the repository being scanned
+                                  [default: local-config]
```
jsimmons1-godaddy, line 95:
```suggestion
                                  [default: target-config]
```

README.md
jsimmons1-godaddy:
```suggestion
                                  setup in the repository or folder being scanned
```

CHANGELOG.md
```diff
@@ -2,8 +2,21 @@ vX.X.X - Mar 3 2023
 --------------------
 
 Features:
+* [#473](https://github.com/godaddy/tartufo/pull/473) - Introduces new flag `-target-config/--no-target-config` to process or skip using config file in the repository being scanned
```
rbailey-godaddy, line 5:
```suggestion
* [#473](https://github.com/godaddy/tartufo/pull/473) - Introduces new flag `--target-config/--no-target-config` to process or skip using config file in the repository being scanned
```

README.md
rbailey-godaddy: Just beating on a dead horse at this point. It's not "config files like tartufo.toml or pyproject.toml", it's exactly "tartufo.toml or pyproject.toml in the target root directory". And I am not necessarily a fan of the phrase "target root directory" and would be happy to see a better term. But it'll never be anything except one of those two files, whichever is found first, in the order listed.

docs/source/configuration.rst
```diff
@@ -8,27 +8,98 @@ unwieldy and lead to an overly cumbersome command. It also becomes difficult to
...
+
+If ``tartufo.toml`` is not present but exists in the parent directory, then the
+effect is that ``../tartufo.toml`` is read first, ``myconfig.toml`` is read next,
+and ``pyproject.toml`` (if it exists) would be processed last because it is in
+the scan target and ``tartufo.toml`` does not exist in the scan root.
+
+Note this behavior is not applicable to remote repository scans, because the
+remote repository will be cloned to a scratch directory and neither that directory
+nor the configuration files within it are available to specify with ``-config``.
+
```
rbailey-godaddy, line 84: On reflection, I think this example would be better if we specified some other target directory that is not `.` and then replace instances of "current directory" with "target" or similar. What I wrote here (from previous PR) is technically correct, but I think a lot of people see "current directory" and think the current directory is magic, without getting the point that it's true here only because the current directory is also the target directory.

rbailey-godaddy (changes requested):
In addition to the twiddly stuff, I think this would be a lot clearer if we could filter out the garbage from the "multiple config" PR that's already been merged. Rebasing (against the squash merge) might be nasty -- can you perhaps just cherry-pick your final commits into a separate clean branch and submit a new PR based on that? Whatever is easiest for you.

emayuri-godaddy:
Closing this PR as it was branched out of a feature branch and rebase seems to be difficult now. 
The new PR for this work  https://github.com/godaddy/tartufo/pull/480

PR #479, opened 2023-03-28
Remove upward traversal logic for config discovery

### LINKED ISSUE #475: Do not traverse directories with --config
## Feature Request

When I say `tartufo --config foo`, tartufo looks for `foo` in the stated location; but if it is not found, it starts moving upward in the directory hierarchy looking for a usable file with the same name.

Tartufo should not perform unrestricted directory traversals looking for configuration files. It should use exactly the file I specify, or use nothing.

sushantmimani:

* [ ] Tests (if applicable)
* [x] Changelog entry
This PR removes the upward traversal logic for configuration discovery in Tartufo. Tartufo will no longer look for configuration files in unspecified parent directories.

## PR Type

* [ ] Bugfix
* [x] Feature
* [ ] Build related changes
* [ ] CI related changes
* [ ] Tests
* [ ] Other

## Backward Compatibility

* [x] No (breaking changes)
This is a breaking change for users who were relying on upward traversal for configuration discovery

## Issue Linking
Fixes #475

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -42,8 +42,7 @@ Consider:
 
    tartufo --config myconfig.toml scan-local-repo .
 
-``tartufo`` will look for ``myconfig.toml`` in the current directory, and then
-in the parent directory if it does not exist in ``.``, etc., and then look for
+``tartufo`` will look for ``myconfig.toml`` in the current directory, and then look for
 either ``tartufo.toml`` or (if not found) ``pyproject.toml`` in the current
 directory (only) because that is the target of the scan. Directives in, say,
 ``tartufo.toml`` would supersede settings in ``myconfig.toml``.
@@ -57,7 +56,7 @@ However:
    tartufo --config tartufo.toml --config myconfig.toml scan-local-repo .
 
 will cause ``tartufo`` to read ``tartufo.toml`` in the current directory
-(assuming it exists, and in parent directories otherwise), and then ``myconfig.toml``
+(assuming it exists), and then ``myconfig.toml``
 as above, and then either ``tartufo.toml`` or ``pyproject.toml`` in the current
 directory (only) because that is the target of the scan.
 
@@ -67,10 +66,7 @@ overriding directives), and ``tartufo.toml`` is not read again because it was
 processed already (and any ``pyproject.toml`` is ignored because ``tartufo.toml``
 was found).
 
-If ``tartufo.toml`` is not present but exists in the parent directory, then the
-effect is that ``../tartufo.toml`` is read first, ``myconfig.toml`` is read next,
-and ``pyproject.toml`` (if it exists) would be processed last because it is in
-the scan target and ``tartufo.toml`` does not exist in the scan root.
+If any file specified by `--config` doesn't exist, tartufo will raise an error and exit.
 
 **Note**: This behavior is not applicable to remote repository scans, because the
 remote repository will be cloned to a scratch directory and neither that directory
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -12,10 +12,10 @@ configuration file!
 
 You can `tell tartufo what config file to use
 <usage.html#cmdoption-tartufo-config>`__ by specifying one or more configuration
-files on the command line. Each file is located and processed in order. When
-conflicting directives are provided in different files, the value in the last
-file processed takes precedence. Paths specified using --config are interpreted
-relative to the current working directory. List-valued directives (such as
+files on the command line. Paths specified using --config are interpreted
+relative to the current working directory.Each file is located and processed in order.
+When conflicting directives are provided in different files, the value in the last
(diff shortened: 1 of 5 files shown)


docs/source/configuration.rst
```diff
@@ -22,13 +22,7 @@ file processed takes precedence. List-valued directives (such as
 Configuration Discovery
 -----------------------
 
-``tartufo`` uses two additional heuristics to locate specified configuration files.
-First, the current working directory is used as a base for relative filenames.
```
jsimmons1-godaddy, line 26: Was this logic removed? It looks to me like only the parent traversal logic was removed, not the logic that tries in the current working directory.

rbailey-godaddy: It's not trying in the current working directory per se. It's saying that pathnames provided with `--config` are interpreted relative to the current working directory.

rbailey-godaddy: Looking at the entirety of the deleted text, it's probably worth adding an explicit statement to that effect (although I suppose it should be obvious). i.e. "Path names specified using `--config` are interpreted relative to the current working directory."

rbailey-godaddy (commented):
One documentation cleanup/clarification point only. Can't argue with the logic, yay!

docs/source/configuration.rst
jsimmons1-godaddy: Apologies, I think I just misread what the original text here was conveying. Looking one more time at the documentation, I don't have any objections at all to this removal.

The examples in https://github.com/sushantmimani/tartufo/blob/remove-upward-traversal/docs/source/configuration.rst#configuration-discovery do need to be updated though, right? They speak at least twice about traversing upwards still.

sushantmimani: Documentation has been updated.

jsimmons1-godaddy (approved):
Looks good to me, for what it's worth. Thanks for making these changes

docs/source/configuration.rst
```diff
@@ -14,21 +14,16 @@ You can `tell tartufo what config file to use
 <usage.html#cmdoption-tartufo-config>`__ by specifying one or more configuration
 files on the command line. Each file is located and processed in order. When
 conflicting directives are provided in different files, the value in the last
-file processed takes precedence. List-valued directives (such as
+file processed takes precedence. Paths specified using --config are interpreted
+relative to the current working directory. List-valued directives (such as
```
rbailey-godaddy: Apologies for being a pain in the ***, but seeing "Paths specified using --config are interpreted relative to the current working directory." in the middle of the discussion about how multiple files are interpreted feels awkwarrd. Can you move this up? (Before "Each file is located...", I think.)

docs/source/configuration.rst
```diff
@@ -14,21 +14,16 @@ You can `tell tartufo what config file to use
...
 -----------------------
 
-``tartufo`` uses two additional heuristics to locate specified configuration files.
-First, the current working directory is used as a base for relative filenames.
-Second, if the specified file does not exist in the specified directory, ``tartufo``
-will search upwards in the directory hierarchy looking for a file with the same
-name.
-
-Additionally, ``tartufo`` will look for a configuration file in the scan target
+``tartufo`` will look for a configuration file in the scan target
```
rbailey-godaddy:
```suggestion
By default, ``tartufo`` will look for a configuration file in the scan target
```

docs/source/configuration.rst
```diff
@@ -48,8 +43,7 @@ Consider:
 
    tartufo --config myconfig.toml scan-local-repo .
 
-``tartufo`` will look for ``myconfig.toml`` in the current directory, and then
-in the parent directory if it does not exist in ``.``, etc., and then look for
+``tartufo`` will look for ``myconfig.toml`` in the current directory, and then look for
 either ``tartufo.toml`` or (if not found) ``pyproject.toml`` in the current
 directory (only) because that is the target of the scan. Directives in, say,
```
rbailey-godaddy: I think the "(only)" in L48 can be removed.

rbailey-godaddy (changes requested):
Sorry, but a little more wordsmithing....

docs/source/configuration.rst
```diff
@@ -12,8 +12,9 @@ configuration file!
 
 You can `tell tartufo what config file to use
 <usage.html#cmdoption-tartufo-config>`__ by specifying one or more configuration
-files on the command line. Each file is located and processed in order. When
-conflicting directives are provided in different files, the value in the last
+files on the command line. Paths specified using --config are interpreted
+relative to the current working directory.Each file is located and processed in order.
```
rbailey-godaddy, line 16:
```suggestion
relative to the current working directory. Each file is located and processed in order.
```

rbailey-godaddy (approved):
I saw a missing space you can add and then I'm good to go!

emayuri-godaddy (approved):

PR #539, opened 2024-10-16
feat: support python 3.13

branchv:

- [ ] Tests (if applicable)
- [ ] Documentation (if applicable)
- [ ] Changelog entry
- [x] A full explanation here in the PR description of the work done

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Tests
- [ ] Other

## Backward Compatibility

- [x] Yes (backward compatible)
- [ ] No (breaking changes)

## What's new?

Just adding support for python 3.13. Poetry was locked to an older version of `pygit2` that can't be built on 3.13 so I had to add a newer version (which no longer supports python 3.8). Also ran `poetry update`


rbailey-godaddy:
Cool! However, there's a lot of breakage in the CI testing... I think it's related to https://github.com/libgit2/pygit2/issues/1317 -- neither our runners nor my workstation are prepared to deal with compiling-from-source in the absence of 3.13-compatible wheels. Hopefully this problem will solve itself in the near future.

branchv:
ah `pygit2==1.16.0` did publish [3.13 wheels](https://pypi.org/project/pygit2/#files) but poetry selected an older version because it also dropped python 3.9 support. I just pushed an update to prefer `^1.16.0` for python>=3.10

tox.ini
rbailey-godaddy, line 1: This seems really stupid, but 3.13 seems to have brought along some breaking deprecations (typing.Pattern now is gone, etc.) that are upsetting mypy and pylint. For the sake of minimizing the work needed to accommodate the change, can you revert the testenv:* changes so we keep using python 3.12 for all of the linters?

Of course if you want to fix the code to placate the linters without breaking backwards compatibility, go for it... But I took a quick stab at it and it looks gorier than I would have thought.

rbailey-godaddy (changes requested):
Stupid linters. :-(

rbailey-godaddy (approved):
Yeah, I'll buy this. (At first I thought I'd accidentally committed some of my testing from yesterday because you'd made many of the exact same changes I considered!)

I see there's a `types-pygit2` library that we could install to address many of the `attr-defined` errors, but it feels like a cure that's worse than the disease -- it doesn't fix all of those errors, and it introduces a number type compatibility errors for known-working code. The resulting vibe I get is "not fully baked".

I think this PR reaches the goal of adding 3.13 support with the smallest amount of perturbation to the codebase. Thanks!

branchv:
Happy to help!

Meant to comment earlier but the `type: ignore` comments were needed because `pygit2` now ships type hints but they aren't 100% accurate yet: https://github.com/libgit2/pygit2/issues/1322 

Also `pylint` complaining about `typing.Pattern` missing was a false positive, that was fixed with 3.13 support in 3.3.0 but similar to pygit2 that version also dropped 3.9 support

emayuri-godaddy (approved):

